PR #140, opened 2015-06-30
Rearranged the keyval/casstring validation to ignore strings

aharjula:
The new implementation for stack_utils::all_substring_strings and some rearranging of the validation to expand the possibilities of keyval and casstring. Feel free to come up a four letter string that breaks this parser...

Maybe there is a reason why strings should not contain @ or $ chars but I could not think of one so I made those chars possible.  Also allows multi-line strings in keyvals...

The code could have been modified to allow "system" as a string but as it was very clearly forbidden in the comments we still block many words in our strings, even though one can easily concatenate oneself around those restrictions... maybe we should just accept that one can construct any string inside the maxima code and therefore we could stop checking for the global forbids inside the strings.

Tried atom as an editor and it seems to have broken some indentations and whitespace... sorry...

-Matti


aharjula:
Line 88 of keyval.class.php seems to have lost an s for some reason...

PR #141, opened 2015-07-30
Fixed the normalization of the parse-tree to be bit less recursive.

aharjula:
Helps with very large CASTEXTs, basically STACK_3.4S-branch could not go forward without this.

Also important for other uses as PHP-recurse limit is typically 100 and if you place hundred or so CAS-variables into to the CASTEXT in certain way, you will cause that recursion to explode.


sangwinc:
Thanks for this commit Matti.  I've pushed a change with the "s" back on line 88 of the keyvals.  Don't know how that got missed off, but well spotted.

PR #146, opened 2015-09-27
Fixed typo in message string

gustavdelius:
Just fixed a simple typo in the message the student gets when the answer is a set with the wrong number of elements.


sangwinc:
Thank you Gustav.

PR #163, opened 2015-11-26
corrected broken link to Lisp implementations

whanrott:
the target link for the Lisp Implementations appears to have been domain squatted. I've pointed it to the SourceForge homepage for Maxima. YMMV as to whether this is the best target. You can check the old target: http://maxima-project.org/wiki/index.php?title=Lisp_implementations


sangwinc:
Thank you.

Issue #162, opened 2015-11-21
Trouble with missing statistics packages


alcarola:
Hello,

We just pulled the latest master branch and noticed that our probability and statistics exercises stopped working. I found [this commit](https://github.com/maths/moodle-qtype_stack/commit/75f4ea0abcd0bde46ead6da3f978b0fd5269f5a2) and uncommented the loading of the three stats packages. This temporarily fixed the problem for us, but are my changes not likely to be overwritten when I update STACK the next time?

One option would of course be to write in the documentation that Debian (and other) users need to make sure they have the statistics packages, but I don't know if this is any good.

Thanks!
Mikael

sangwinc:
Thanks Mikael,

I'm sorry about this.  Commenting out the loads in the code are only intended as a temporary measure.  Without this, people couldn't install STACK at all. Thanks for opening an issue.

I can't figure out a way to autodetect this problem. I have tried.  I think we may have to add another option, which again makes install problematic.  Does anyone have suggestions about this?

Chris

sangwinc:
Mikael,

I've added an additional option to let users load optional Maxima packages.  The default is to load the packages which are causing a problem, but to load them in such a way that failure does not bring STACK to a grinding halt.  This mechanism is extensible.  I've been asked before for support for various packages, and this will provide one mechanism to support this as well.  Comments on this approach are very welcome indeed.

Chris

alcarola:
This sounds like a good solution to me! In a while we will probably 
switch from the master branch to the abacus branch. I assume the fix 
will eventually find its way into that branch too.

aharjula:
Indeed, the abacus branch will get a merge of most of masters latest commits before christmas.

Now that we have installation specific configurations of packages we may need to think about adding metadata to questions about the packages they need and some sort of a way to tell the importer of questions that they will not work without extra packages... But there is obviously the issue of finding out what packages the questions need, maybe just add the list of packages configured on the installation exporting and comparing it to the situation at the importing side and warning if different.

Also I would like to point out that those specific packages now listed include some of the slowest loading packages and may require pre-loading and as the loading now happens on separate lines outside stackmaxima.mac this may cause some issues if people do not adapt the normal optimisation guides accordingly. Basically, these too will need to be included in the memory images. Anyway, it was good that you chose not to do this through live connections as that would made optimisation with memory-images very difficult.

sangwinc:
Thank you Matti,

[...]

PR #165, opened 2015-12-11
Moved the "STACK-Maxima started" printing to the end of loading of ad…

aharjula:
…ditional libraries.

This is for the issue #162.

### DIFF
--- a/db/install.php
+++ b/db/install.php
@@ -26,7 +26,7 @@ function xmldb_qtype_stack_install() {
     global $CFG;
 
     // Define stackmaximaversion config parameter.
-    if (!preg_match('~\[ STACK-Maxima started, library version (\d{10}) \]~',
+    if (!preg_match('~stackmaximaversion:(\d{10})~',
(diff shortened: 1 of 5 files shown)


aharjula:
Argh... there are plenty of install and upgrade related scripts that try to find that string from stackmaxima.mac... so wait for another version.

timhunt:
Why not get it to output an additional string, like "All libraries loaded"? Then you don't have to change the existing strings.

aharjula:
In that case we would need to warn all the users of MaximaPool about the change in the logic of the "STACK-Maxima started" string. As it no longer marks the point where the maxima is actually ready for use. Somehow I would like to keep that string sensible and meaning what it seems to mean instead of minimising code changes and adding quirks. 

Basically, I read that string as a promise and we are not keeping that promise but we should and we can. But if we intend to not keep that promise we should just remove it from the world, which would again cause issues on the Pool side.

timhunt:
The patches look OK to me (apart from some whitespace that is not quite right.)

sangwinc:
Thanks Matti.

PR #188, opened 2016-02-19
clarified documentation about optimising maxima

whanrott:
I found that the statistics packages weren't loaded. This is because I had saved an image without the packages loaded. When I loaded them the healthcheck complained about incompatibility. The fix is to load all the required libraries. Tested with maxima 5.36.0 built from source.


sangwinc:
Thank you very much indeed for this helpful correction.

PR #189, opened 2016-02-19
clarified descriptive text for qtype_stack | maximalibraries in qtype_stack.php

whanrott:
added supporting text pointing out that an optimised image of maxima would not pick up libraries entered in the field "Load optional Maxima libraries:" (qtype_stack | maximalibraries )


sangwinc:
Thank you very much indeed for this helpful correction.

PR #190, opened 2016-02-23
Replace \ with / in paths

christianp:
You never know who might get tripped up by the wrong kind of slash.


sangwinc:
Thanks Christian,
Chris

PR #204, opened 2016-04-04
added build notes for maxima on Ubuntu

whanrott:
I built a fresh instance of Ubuntu 14.04.04 LTS using Vagrant. Then applied this script. It compiled without errors.


sangwinc:
Thank you very much for reporting this.  Very helpful to know.

PR #212, opened 2016-06-09
Fix #203, by handling strings before array splits and doing some unescaping.

### LINKED ISSUE #203: Math and text in MCQ
I'm trying to make use of a [MCQ](https://github.com/maths/moodle-qtype_stack/blob/master/doc/en/Authoring/MCQ.md) as part of a question. The goal is to make sure the student chooses the right probability distribution for the rest of the question. So far I have something like this:
```
AnsList: [[1, false, "Normal distribution"], [2, true, \chi^2]];
```

The problem I run in to is that I can't combine math and text in the display field. I would like to have something like `[2, true, \chi^2\text{-distribution}]` or `[2, true, \(\chi^2\)-distribution]` but then I just get error messages like `CAS failed to return any evaluated expressions. Please check your connection with the CAS.`

Is it possible to combine text and math in the same field?

aharjula:
Chris, your fix for this does things differently, but I needed this now and do believe this causes way less complications...

### DIFF
--- a/stack/input/dropdown/dropdown.class.php
+++ b/stack/input/dropdown/dropdown.class.php
@@ -111,7 +111,15 @@ class stack_dropdown_input extends stack_input {
          * display - the LaTeX displayed value.
          * correct - whether this is considered correct or not.  This is a PHP boolean.
         */
-        $values = stack_utils::list_to_array($teacheranswer, false);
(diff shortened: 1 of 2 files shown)


sangwinc:
Thanks Matti, this looks good to me.

PR #213, opened 2016-06-09
Always filter equation validation

tlock:
This item implements a fix that shows the validation of the equation correctly in realtime without the need for clicking the 'Check' button to do the validation.

### DIFF
--- a/stack/input/inputbase.class.php
+++ b/stack/input/inputbase.class.php
@@ -487,7 +487,7 @@ abstract class stack_input {
             return '';
         }
         $feedback  = '';
-        $feedback .= html_writer::tag('p', stack_string('studentValidation_yourLastAnswer', $state->contentsdisplayed));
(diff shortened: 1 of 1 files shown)


sangwinc:
Many thanks indeed.

timhunt:
This change is wrong. It was causing double-filtering. (stack_string does filtering.)

I am about to revert this as part of fixing the unit tests to work with Moodle 3.1.2.

Tim, why did you think this was necessary? Neither your bug report, nor commit comment, gives any explanation.

tlock:
Hi Tim,

If you enter in an answer to a stack question, the text isn't translated till you click the "Check" button in version 2015090902. Has that been fixed otherwise?

timhunt:
Before reverting your change, I checked that the input validation was being rendered correctly, and it was. So, I guess that is fixed. Thanks for clarifying.

PR #219, opened 2016-06-28
Abacus 1.2

aharjula:


aharjula:
Argh, this style checking stuff and travis. I'll do some cleaning before the acepting then.

Issue #225, opened 2016-08-04
Forbidden words inside strings?


aharjula:
Lately we have had the pleasure of working with the MCQ-inputs and as those allow you to give labels as strings we have been able to do interesting things, but as the casstring validation checks the thing for forbidden words without caring about where the words appear things have been slightly difficult in some cases.

For quite some time we have come across the issue of having strings with words like "system" in them and we have been able to work around that by simply concatenating the difficult words from smaller parts, but it is inconvenient and makes the code less readable. So could we maybe change the bad word checking so that it would not check inside strings? Also could we do it so that when PRTs check references to inputs they would not check inside strings (somewhat related to #194 )...

So does anyone have any objections to changing [casstring->check_security](https://github.com/maths/moodle-qtype_stack/blob/master/stack/cas/casstring.class.php#L918) so that it would strip strings out before searching for the bad words? Should not have any actual security effect as we can already use concatenation to get those bad words past it as strings, this would only allow us to get the strings in without concatenation...

sangwinc:
Matti,

I've just been very very cautious here!  If you can really be sure there is no way to turn a Maxima string into a command, then I've no objection.  I know this is going to be a problem with words like "system" inside strings...

aharjula:
I can't be sure that I can block turning string into a command, but as I said you can already get those evil words through the validation by simply concatenating them, so as long as the validation blocks the commands outside strings that could cause evaluation of strings containing evil things everything should be safe even if we make it less difficult to input strings that have common words like "time" or "error".

sangwinc:
Yes, you are right Matti.  
This is also a worry....
Would you be willing to create a branch with a concrete proposal.  I don't object to your suggestion, but have just been overly cautious in the past.
Chris

PR #234, opened 2016-09-01
Allow forbidden words inside strings, #225

aharjula:
Remove safety measure that does not protect from the thing it tries to protect from but causes difficulties for new use-cases. Issue #225


sangwinc:
Thanks Matti,
Chris

PR #236, opened 2016-09-03
update maxima.md, remove duplicate exdowncase(ex)

gitnovisat5:
function exdowncase(ex) is/ was mentioned twice


sangwinc:
Many thanks - all help like this is much appreciated.

PR #285, opened 2017-03-16
Explain what the Diff and Int answer tests do

christianp:
It wasn't at all clear from the docs what the "Diff" answer test does, so I've investigated and then had a go at expanding the relevant section in the docs.

I also added a bit to the "Int" section, because that doesn't actually say how the test is passed.


sangwinc:
Thank you Christian, very helpful.

PR #303, opened 2017-08-18
Forgotten abacus legacy function and maybe direct link to material conversion.

aharjula:


aharjula:
Will merge myself, just doing Travis check.

PR #320, opened 2017-10-26
Always show question parts in the same order

timhunt:
Before this change, on the Edit question form, inputs (and PRTs) were
shown in the order they appeared in the question text, while on the
Question tests page, they appeared in alphabetical order.

Not a terrible thing, but nicer to keep it consistent.

It seemed safest to get the order right when the question is
intitialised, even though this is a small amount of unnecessary work
in some cases. I don't think it is significant.


timhunt:
I must admit, I have not run all the tests yet. I was being lazy and hoping Travis would do it for me. It has, at least, started.

Issue #327, opened 2017-12-03
Create Maxima Image fails because of config caching.


sgparry:
During the "Create Maxima Image" process, Stack changes between different Maxima configurations, first to remake the image, then to test the new image and, where needed, to revert to original settings.
This fails to work because the config is cached in the utils class.

sgparry:
This is, I think, now fixed. Can someone please merge the pull request so that it can be closed? Thanks.

sangwinc:
Thanks for pushing this.

Please let me apologize about the delay here.  I'm just back from Oxford.  Someone spilt coffee over my laptop on Monday, so I've been out of commission for a couple of days.  Things have backed up, and I'll certainly look at this over the next couple of days.

sgparry:
Sorry to hear about the laptop. :-(
It's OK ; it gave me a chance to get the fix right. I only just realised something had screwed up some of the source line endings to CRLF (yuck).

sangwinc:
Thank you very much indeed for this contribution.  Much appreciated.

PR #328, opened 2017-12-03
Fix for issue #327

sgparry:
Added lines to update the config cache to match the db.


sgparry:
Modified the commit for compatibility with PHP5 and because I had only succeeded in fixing half the problem. Now the maximalocal.mac file is generated correctly. Also, errors are better reported and

PR #338, opened 2017-12-29
Remove libs from SBCL instructions

dthies:
These changes have been tested in CENTOS 6/7 and seem to resolve the stack page exception problem that appeared.


sangwinc:
Thank you Daniel,

Chris

PR #354, opened 2018-03-05
Fix for iss353 - introduced new function stack_kill() as workaround.

sgparry:


sangwinc:
Thank you very much.  Sorry again about the conflict with "all".....

PR #366, opened 2018-04-06
Guidelines

aharjula:
Some guidelines that might be useful for other STACK users might be nice to include in the STACK documentation.

I don't think I will come up with anything more for this so maybe it could be merged so that others may give more comments.


sangwinc:
Great idea, thanks for adding this Matti,

PR #372, opened 2018-05-09
Improve input usability on iPads, etc.

timhunt:
I was demoing STACK on an iPad yesterday, and it was a real pain because if you tried to type x + 1, it automatically changed it to X + 1. (It upper-cases the first character of input by default.)

Fortunately, HTML5 provides a way to tame this behaviour, which is what I implement here. See https://caniuse.com/#search=spellcheck and https://github.com/Fyrd/caniuse/issues/1729.

However, I have not yet tested this on an iPad to verify that it acutally has the desired effect, and it is too late this evening. I just wanted to get the code written. I have run the unit tests, and they pass.


sangwinc:
Thanks Tim,
This has the very pleasant side effect of not underlining maths with a red squiggle to indicate a spelling mistake! I didn't know that was possible.
Chris

Issue #373, opened 2018-05-21
PHP time-outs when clearing the CAS cache


timhunt:
We have a busy Moodle site that uses STACK a lot. Therefore, our CAS cache, and the corresponding cache of plot files in Moodle data can get very large. (We currently have 200,154 plot files on disc.)

Therefore, sometimes clearing the CAS chache times out, and when this happens during a Moodle upgrade, it si not only a big pain, but also, it was meaning that stackmaximaversion was not getting updated in the config_plugins table.

Also, clearing the cache was only deleting PNG plots. It was missing SVG plots.

PR #374, opened 2018-05-21
Avoid time-outs and the consequences when clearing the CAS cache #373

timhunt:
While clearing the cache, we show a progressbar and call
core_php_time_limit::raise periodically. (This requires an extra
Continue click after the the cache has been purged using the button on
the Healthcheck page, but this actually makes it clearer that the cache
purge succeeded, so might be a good thing.

Clearing the cache now also deletes SVG plots.

During upgrade, we set_config('stackmaximaversion', ...) before clearing
the cache, because it is safter that way. (You can manually clear the
cache later, but you can't easily update that config value other than
directly in the database, if something goes wrong.


sangwinc:
Thanks Tim,
As STACK gets larger we will have this kind of thing.  Thanks for fixing it.

PR #379, opened 2018-06-05
Api v4

mloginov:
Added codemirror mode and hint for yaml question syntax


sangwinc:
Thank you very much indeed for this contribution.

PR #388, opened 2018-07-26
Revert "Authoring_quick_start_2.md file translated into Spanish"

sangwinc:
Reverts maths/moodle-qtype_stack#382


sangwinc:
This should not be pushed into "master" just yet!

PR #393, opened 2018-07-31
fix two typos for languages

germanvaleroelizondo:
I think I saw these two tiny typos in the new English language strings for the new plugin version 3.5.


sangwinc:
Thank you very much indeed.
Chris

Issue #398, opened 2018-08-16
Version check broken for Maxima Nightlies on Ubuntu


potslinux:
One possible way to install maxima on Ubuntu is by using the maixma developers nightly builds, s.
https://launchpad.net/~maxima-developers

But the STACK plugin fails in this case. TMHO, the problem lies in the version check, e.g. in stack/maxima/stackmaxima.mac:
  55   MAXIMA_VERSION_STR: ?\*autoconf\-version\*,
  56   MAXIMA_VERSION: map(parse_string, tokens(?\*autoconf\-version\*, 'digitcharp)),
  57   /* We need the "if" statement below, because versions compiled from source give erroneous results. */
  58   MAXIMA_VERSION_NUM: float(MAXIMA_VERSION[2]+(if is(MAXIMA_VERSION[3]<10) then MAXIMA_VERSION[3]/10 else 0)),

The nightly build do not have version strings with three mumbers (5.41.0) but something like "5.41post". The Version check than fails in line 58 with an exception because of invalid array field in MAXIMA_VERSION. Maybe the version check could be adopted to this "invalid" version strings.

My environment:
Ubuntu 16.04.05 LTS
Moodle version 2018051701.05, release "3.5.1+ (Build: 20180803)"
qtype_stack version  2018073000, release "4.2 for Moodle 3.0+"
Maxima 5.41.0-27~201808090736~ubuntu16.04.1 installed via PPA 
"http://ppa.launchpad.net/maxima-developers/maxima-nightly/ubuntu xenial main"

sangwinc:
Ok, well the intention wasn't to break things for nightly build people (but I can offer limited support for random versions of Maxima!).   Can you propose a change as a pull request and I'll review it?
Thanks,
Chris

potslinux:
I created a pull request or at least tried to do so. There are only two lines to change.

sangwinc:
Many thanks for that contribution. Much appreciated.

PR #400, opened 2018-08-20
Maxima nightlies

potslinux:
Calculation of MAXIMA_VERSION_NUM improved to avoid problems with nightly builds of maxima by ubuntu maintainers, s. issue #398.


sangwinc:
Many thanks indeed.

PR #404, opened 2018-09-04
Accessible: Add answer label for input

trieuhaiquan:
This is an accessibility bug. 
As all input of the other question has a <label> tag, so this should be added following the pattern used by other Moodle question types.

### DIFF
--- a/lang/en/qtype_stack.php
+++ b/lang/en/qtype_stack.php
@@ -47,6 +47,7 @@ $string['stackversionmulerror'] = 'This question has an input which uses the "mu
 // Strings used on the editing form.
 $string['addanothernode'] = 'Add another node';
 $string['allnodefeedbackmustusethesameformat'] = 'All the feedback for all the nodes in a PRT must use the same text format.';
+$string['answer'] = 'Answer: {$a}';
 $string['answernote'] = 'Answer note';
 $string['answernote_err'] = 'Answer notes may not contain the character |.  This character is inserted by STACK and is later used to split answer notes automatically.';
 $string['answernote_help'] = 'This is a tag which is key for reporting purposes.  It is designed to record the unique path through the tree, and the outcome of each answer test.  This is automatically generated, but can be changed to something meaningful.';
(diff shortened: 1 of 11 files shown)


timhunt:
I agree with Matti. This does not follow the pattern set by other question types (e.g. qtype_shortanswer with the input box in the middle of question_text using ________, or qtype_cloze, or, in the plugins database qtype_combined).

There is no way, in general, to deduce the relevant part of the question text to mark as the label (in Matti's example the "f(x) = "). Therefore, the lable woudl have to be artificial.

If it is going to have to be an artificial label (which is what the other qtypes do to tick the accessibility box) then the lable has to be different for each input. Adding an identical lable of 'Answer' to each input does not help anyone.

timhunt:
I don't know the best answer to this. There is lots of online guidance for marking up form elements, but it all seems assume that you are marking up a form, not accepting input in context inside other content. (I tweeted https://twitter.com/tim_hunt/status/1048155481246781441 to see if that got me any advice.)


1. So, in terms of the actual lable, then I cannot think of anything better than what has been done here: "answer 1", "answer 2". However, it would be better without the colon.

We need to do something more specific to label the multiple input boxes for matrix inputs.

And, it would be better added using an aria-label attribute, rather than being added inline in the HTML content flow.


That first point is where we started, but as Chris says, there are other issues:


2. There should be some sort of link from the input to the corresponding validation. I don't think that using the vaidation as a label for the input is right.

Certainly the validation area should have aria-live="polite" added.

I would be interested to see if using aria-describedby to link the input to the validation area does any good.

It would probably also be good to add <h5 class="accesshide">Validation for answer input {N}</h5> at the start of the valdiation informtion generated by stack_input::render_validation.


3. On a purely technical point, if you are working on question/type/stack/yui/src/input/js/input.js, is now the time to convert it from YUI to AMD? That would get rid of the last YUI use in STACK, but really that is a separate issue.


Overall, I think we are not going to solve this well without getting some input from real screen-reader users. It is also exprient yourself with the https://www.nvaccess.org/download/ or the free trial of Jaws, but that only goes so far. Screen-readers are specialist tools, and we are only ever going to be amateur users.

That is the extend of my non-expert knowledge. I will say that, reading through a lot of https://www.w3.org/TR/WCAG20-TECHS/Overview.html this morning, I think STACK is getting a lot of things right. (e.g. putting the validation near the input, having good validation, ...)

sangwinc:
This is a very important issue, and we need to get it right.  So, we should try to find someone with expertise in both the mathematics and the software.

Issue #412, opened 2018-10-26
cURL add support for SSL options


adpe:
Hello @sangwinc 

For run the MaximaPool over secure mode (SSL) you must give some parameters, that cURL can successfully verify the certificate.

Something similar I found in the `.../moodle/search/engine/solr` setup. Here's the configuration for the connection:
![image](https://user-images.githubusercontent.com/9467708/47548558-c9dd9b00-d8f9-11e8-9192-d49573209c1e.png)

At the moment for my case it's enough to just add a new settings field like:
![image](https://user-images.githubusercontent.com/9467708/47548668-f5f91c00-d8f9-11e8-887b-4c1db5a4e37e.png)

But you must be aware, that when you uncheck this field, it can be a risk. But what's more risky? HTTP or HTTPS without certificate validation? If we want to check the certificate we must add more cURL options to get it working.

What do you thought about this case?
Adrian

aharjula:
Well, one might also think about why one needs HTTPS for MaximaPool at all? Essentially, the Pool should never be directly exposed to the public Internet and the only connections it should receive should be through internal networks and even in those cases only from the Moodle/ILIAS servers running STACK (+ maybe some admin users). Within such network setup it should not matter whether the information in transit is readable or even modifiable as you are working in a trusted context.

If one needs/wants to run SSL then one can simply modify the STACK side connector-class to set those settings for cURL. But personally, I would consider SSL as wasted resources for MaximaPools in internal networks and firewall controlled access policies. But if you add some form of authentication to your Pool and place it into a network where unknown people can see and access it then you might need that, however if you go for such a setup you are essentially asking for someone to crack it.

Also SSL handshakes take time and MaximaPool exists to save time, so if possible keep unnecessary time use to a minimum and skip SSL if at all possible.

adpe:
Hi @aharjula 

Yes that's right. In most cases it makes not much sense, but there can be individual cases. I've seen, that with SSL the performance is lower.

I'll look how to deal with it and our `Security officer`.

Nevertheless, I will publish my solution.

timhunt:
I don't understand why the Solr config screen requires so many millions of settings. HTTPS URLS should just work in most places.

If you are setting up https, why not use a proper certificate?

If we consider it essential to make self-signed certificates work, the a single option should be enough. However, I think the config variable should have its meaning inverted, so it fails safe. E.g. call it casskipmaximaservercertverfiy.

[...]

PR #413, opened 2018-10-26
Checkbox to deactivate SSL certificate verify (#412)

adpe:


adpe:
As mentioned here https://github.com/maths/moodle-qtype_stack/issues/412#issuecomment-597022731 it will be closed.

PR #414, opened 2018-11-01
Command-line script to clear the CAS cache

timhunt:
This is something we want at the OU to help automate our releases.

Obviously, the cache is purged automatically as part of the STACK upgrade. However, our CAS cache can get very large (e.g. 350,000 plot files in Moodledata) and deleting them all takes many tens of minutes. We would like the option to purge the chache in advance, so we don't have to wait all those minutes while Moodle is in maintenance mode. Hence this script.

### DIFF
--- a/cli/clearcascache.php
+++ b/cli/clearcascache.php
@@ -1,0 +1,53 @@
+<?php
+// This file is part of Stack - http://stack.maths.ed.ac.uk/
+//
+// Stack is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Stack is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Stack.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Script to clear the CAS cache from the command-line.
+ *
+ * @package    qtype_stack
+ * @subpackage cli
+ * @copyright  2018 The Open University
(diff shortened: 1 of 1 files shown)


aharjula:
350k plot files in a dir, that must be bad... Imagine if we had even one active script running a directory listing on that. Have you had any issues with the filesystem you use for Moodledata when that amount of files builds up? If not what filesystem is it? NFS or some such but what is behind that?

Could we maybe also think about adding some form of sub directory structure into the plot dir, to spread out those files to directories with less files. Back in the STACK 1 days when I rendered every bit of castext as bitmaps I used a three level structure where each level only had 256 dirs I used the six first chars of the images name for defining into which of those directories it belonged and as those names were hex-form hashes that balanced itself quite fine. If I remember correctly the worst case we had was that we had about five images in the same dir. Why I did that was due to keeping them in a single dir broke the filesystem and for somewhat obvious reasons access times grew, not to mention calling `ls` in such a dir was a suicide. However, I never had quite that many plots to handle so they were never a problem for me, but if you have 350k plot files it might be of interest to check if that affects access times and would merit for some action to be taken. I.e. you might want to check the time it takes to serve a random plot from that directory if there is a few plots there as opposed to having 350k plots...

timhunt:
Yes, some sort of sharding of the directories might be good.

I believe that in the plot filename (e.g. stackplot-3750071431-46248239.svg) the first number is the current timestamp (Maxima code string(absolute_real_time())) and the second number is string(rand(10^8)).

So, the second number will be randomly distributed, and so a good hash key. It is proably enough to put the plot in a directory named for the first 3 digits of that, then we would have 1000 folders each containing about 4000 files. that should be fine.

However, really what we should do is change the code so the mapping plot filename <-> path on disc is just defined in two functions (one to go each way) so it is easy to change in future. (I don't currently have a lot of time to work on that, so I am happy for anyone else to have a go.)

timhunt:
I just made Issue #415 to further discus the where to store files issue.

sangwinc:
Thanks Tim, very sensible contribution.

PR #423, opened 2018-12-03
Api v4

vesal:
Changes so that install.php is run after the container has started. Is is to avoid
build problems that comes from maxima violating seccomp rules.


sangwinc:
Sorry, we don't want to merge this into "master".  I think it is a bit early for that...

PR #424, opened 2018-12-03
run install.php after container is running to avoid seccomp problems …

vesal:
…when building the Docker image


sangwinc:
Thank you very much indeed.

PR #428, opened 2018-12-13
Api v4

vesal:
- attribute verifyvar for endpoint.php to make just input verification for on input field
- attribute ploturl for endpoint.php to tell the image URL prefix


sangwinc:
Sorry Vesa,
You want to commit to master here!  We want to keep all this in the api_v4 branch for now.

PR #429, opened 2018-12-13
Api v4

vesal:
- ploturl = for telling what si the real plot server
- verifyvar = tell what input field to verify


sangwinc:
Ok, now I see what you have in mind!

I'm really sorry, but the "verification" is not as simple as this.  The STACK validity checks actually need the whole question.  There are lots of options in validation, including things like checking the student's answer is "the same type" as the answer of the teacher.  So, this includes if the teacher's answer is a set/list/equation/inequality/matrix/expression.  Without the full question you can't do this.  

This is a very neat idea, but it doesn't work!  Sorry....  

Chris

sangwinc:
Thank you for these changes.

PR #430, opened 2018-12-18
Community

birdanja:
Since @sangwinc suggested at the [STACK Conference 2018](https://www.stack-konferenz.de/) to use github also for community building, I looked for the info there on community. I couldn’t find the community.md right away, that’s why I started editing in the readme.md, but moved that to community.md later.

Do we want to keep the references to the STACK-Plugin for ILIAS in the FAQ? What do you think about further references between “STACK for moodle” and “STACK for ILIAS”? (@fneumann, @jcopado)

Could we make “Community” more prominent? As mentioned, it took me a while to find it. I would mention it somewhere in the readme.


sangwinc:
Thank you very much indeed.  Contributions like this are very welcome.

PR #437, opened 2019-01-18
Hide the 'Question tests' link in iframes

timhunt:
Whe a STACK question is in an iframe, which might happen if you are
using https://moodle.org/plugins/filter_embedquestion, then it is a reall bad user-experience if you
can navigate to the 'Question tests & deployed versions' page.
Therefore, we hide the link in that case.

For embedded questions, authors will have to access the question tests
via the question bank.

(We just did a similar change in our 'Pattern-match' question type, and I thought I should submit the same change to STACK.)

### DIFF
--- a/styles.css
+++ b/styles.css
@@ -55,6 +55,10 @@
     font-size: 0.7em;
     margin: -1.3em -3em;
 }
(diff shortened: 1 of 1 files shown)


sangwinc:
Thanks Tim,
Chris

PR #450, opened 2019-03-26
Adding SVG icon.

lucaboesch:
Do you fancy a SVG icon (for question type choser and question listing)?
This is what I would imagine it would look like rendered as SVG.
Best,
Luca


sangwinc:
Thank you so much for contributing this, it looks great!

I think an SVG would be great, and long overdue.

PR #456, opened 2019-04-14
Iss317 - base N support (finally!)

sgparry:
OK, this is not the finished article, but I have been using it for a while now and it serves a purpose; here are the areas requiring further development, along with more testing:

1. The 'interpreted as' output in answer validation is pretty counter-intuitive, as it shows the function call wrapper (frombasen) used to process the students' answer.
2. Matrix handling; I have not fully tested whether the base N functions can be used in a matrix context and whether there is a way of making matrices more null friendly. This is important to me because it would allow the student enter working more easily and me to actually mark it.
3. Library functions; in presenting effective questions and answers I use a lorry load of helper definitions that do things like converting to from floating point binary format, producing levelled random questions (e.g. based on bit count) as so on. Currently I am just pasting them in the question box inline, but many could be put together as another library.

### DIFF
--- a/doc/en/CAS/Numbers.md
+++ b/doc/en/CAS/Numbers.md
@@ -128,14 +128,14 @@ The following commands which are relevant to manipulation of numbers are defined
 
 The following commands generate displayed forms of numbers.  These will not be manipulated further automatically, so you will need to use these at the last moment, e.g. only when generating the teacher's answer etc.
 
-| Command                         | Description
-| ------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
-| `dispdp(x,n)`                   | Truncate \(x\) to \(n\) decimal places and display with trailing digits.  Note, this always prints as a float (or integer), and not in scientific notation.
-| `dispsf(x,n)`                   | Truncate \(x\) to \(n\) significant figures and display with trailing digits.  Note, this always prints as a float, and not in scientific notation.
-| `scientific_notation(x,n)`      | Write \(x\) in the form \(m10^e\).   Only works reliably with `simp:false` (e.g. try 9000).  The optional second argument applies `displaysci(m,n)` to the mantissa to control the display of trailing zeros.
-| `displaydp(x,n)`                | An intert internal function to record that \(x\) should be displayed to \(n\) decimal places with trailing digits.  This function does no rounding.
-| `displaysci(x,n,expo)`          | An intert internal function to record that \(x\) should be displayed to \(n\) decimal places with trailing digits, in scientific notation.  E.g. \(x\times 10^{expo}\).
-| `basen(x,n[,mode[,mindigits]])` | An inert internal function to record that \(x\) should be displayed and input in a base\(n\) integer format with at least mindigits digits.
+| Command                                         | Description
+| ------------------------------------------------| ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+| `dispdp(x,n)`                                   | Truncate \(x\) to \(n\) decimal places and display with trailing digits.  Note, this always prints as a float (or integer), and not in scientific notation.
+| `dispsf(x,n)`                                   | Truncate \(x\) to \(n\) significant figures and display with trailing digits.  Note, this always prints as a float, and not in scientific notation.
+| `scientific_notation(x,n)`                      | Write \(x\) in the form \(m10^e\).   Only works reliably with `simp:false` (e.g. try 9000).  The optional second argument applies `displaysci(m,n)` to the mantissa to control the display of trailing zeros.
+| `displaydp(x,n)`                                | An intert internal function to record that \(x\) should be displayed to \(n\) decimal places with trailing digits.  This function does no rounding.
+| `displaysci(x,n,expo)`                          | An intert internal function to record that \(x\) should be displayed to \(n\) decimal places with trailing digits, in scientific notation.  E.g. \(x\times 10^{expo}\).
+| `basen(x,n[,mode[,inpdigits[,mindispdigits]]])` | An inert internal function to record that \(x\) should be displayed with at least mindispdigits and input  with exactly inpdigits digits in a base\(n\) integer format.
 
 
 | Function                  | Predicate
--- a/stack/basenoptions.class.php
+++ b/stack/basenoptions.class.php
@@ -124,7 +124,7 @@ class stack_basen_options {
     }
 
     private static function digit_range_pattern($radix, $underscoresallowed) {
-        $r = $radix - 1;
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -0,0 +1,2 @@
+/thirdparty/php-peg/nbproject/
```
aharjula: Can I interpret this .gitignore to mean that you had php-peg in use at some point but not anymore? If so then things are all fine, I would like to drop php-peg when the CASText parser gets rewritten and the Maxima-parser comes into play. Instead of php-peg my current view is that PEG.js is good and using it to define parsers for both editor components and back-end components is a big benefit, that php-peg just can't match.

sgparry: I have not had cause to use any of the php-peg or PEG.js stuff in this PR - the .gitignore entry is specifically related to some netbeans related pollution that crept in to the source at some point from who knows where. I am not even sure it was my addition.

Currently, my base-n parsing is very, very low rent - it just uses some regexps to do the job. Obviously, using a generated parser would make much more sense; I use ANTLR myself extensively in other contexts and have experience of good ol' Flex+Bison. However, I would not want to do that in STACK in isolation - moving to a proper collection of parsers is something we should do as a unified effort. Having two different parser generators would be pretty self-destructive.

aharjula: Ahh "nb" as in netbeans as opposed to number base... ANTLR would naturally allow proper parser generation but because I never got the chance to go to a compiler course I never learned that so PEG-parsers are about what I can do, and currently the STACK development pipeline already has Maxima statement parsing coming in the form of a PEG.js parser and the CASText2-project which essentially compiles CASText to single pass Maxima expressions is coming pretty soon after it.

Having now looked at what you have done there are some places where the parser based casstring system will conflict greatly but once you get to act on an AST presentation of the students input it should be easier to handle all those conversions to `frombasen`, it should even be possible to directly give those basen_options to the Maxima-parser so that it would directly identify and even convert/wrap matching tokens.

sgparry: PEG in itself looks interesting - I think you would have found ANTLR just as easy / easier, but the PHP and JS targets are still very alpha. The only thing that concerns me is why do the parsing in JS rather than PHP ? I inherently mistrust anything that relies on code that executes in the browser...

aharjula: Well PEG.js is a parser generator like ANTLR, although only for PEG-style grammars. And it is used to generate a PHP parser that seems somewhat cleaner than the PHP-peg one, it can also generate, from the same grammar, JS and TS parsers that are quite handy when you build editors that do run on the browser as you can then provide instant validation at the browser side without having to send things to the server, the same validation naturally happens there too but responsiveness is nice to have. So I would not worry about running stuff in browser, nothing that gets validated by that browser side parser ever gets through without also being validated by the backend side parser which for now is PHP but could probably be something else if we ever were to port STACK to a non PHP-backend.

sgparry: Ah! That sounds just the ticket, cool. The one thing to bear in mind with grammars applicable to base N is that there is not one grammar, but several, depending on the style of notation used. The code I have written supports Suffix, plain non-greedy and greedy, C++/Java and Visual Basic.

sgparry: I am getting a lot of Travis failures outside of my code. Is that my changes fault or someone else's?

aharjula: Indeed many ways of doing this, in my current AST presentation we have Integers, Floats and Identifiers as suitable tokens for this and it would be simple to just expand the definition of any or all of them to accept a superset of all those notations and then just check them in post processing for the correct syntax. But some of those do have certain side-effects that will surely need some thought in the parser and therefore multiple parsers or a parser with configuration parameters is the the way to go.

Luckily `&` is not an operator in basic Maxima and thus that BASIC syntax is not a problem, the `0x..` and `0h..` style variants should be easy, and greed is easy from the parsers point of view as it is already greedy. The problem will probably be with the suffix notation but that does not seem too bad. The parser just needs to have enough options to cover all those to cover all things.

Just have to be careful as something like `true_36` might give nasty surprises somewhere.

Am I correct to assume that we do not have floats in different number bases? Are they coming?

aharjula: As to the Travis `(kg*m^2)/s^2` style issues are normal, but there seems to be some others there also maybe related to someone else's work but that is difficult to say. In any case I would not worry as I doubt Chris is going to merge this just now anyway as the MaximaParser branch is probably the next major merge (probably happens early next month) and it would not make sense to merge this twice in so short a time.

sgparry: Within my remit at least, floats are expressed as exponent and mantissa in binary integer format, with an implied, invisible binary point (This is a deliberate thing on the part of our exam board); so there are no syntax implications as such. There are some fairly involved functions and expressions to do the denary float to binary float conversions, and I was considering packaging them up for submission as a library.

stack/cas/installhelper.class.php
```diff
@@ -29,7 +29,7 @@ class stack_cas_configuration {
     protected static $instance = null;
 
     /** @var array This variable controls which optional packages are supported by STACK. */
-    public static $maximalibraries = array('stats', 'distrib', 'descriptive', 'simplex');
+    public static $maximalibraries = array('stats', 'distrib', 'descriptive', 'simplex', 'bitwise');
```
sangwinc: Is needed for the base n work, or just another useful library of Maxima functions you'd like the option of having?!

I've reviewed this, and it looks like a sensible inclusion, so I'm happy to have it here.  If the base n work needs it then we should just load it automatically, otherwise it will cause confusion if people don't have this, and then things break.

sgparry:
Bitwise is not needed for Base N per se - just needed by me in setting questions based on bitwise ops on binary numbers. So I would say put in the supported list but don't load it by default.

On 21 April 2019 17:24:45 BST, Chris Sangwin <notifications@github.com> wrote:
>sangwinc commented on this pull request.
>
>
>
>> @@ -29,7 +29,7 @@ class stack_cas_configuration {
>     protected static $instance = null;
> 
>/** @var array This variable controls which optional packages are
>supported by STACK. */
>-    public static $maximalibraries = array('stats', 'distrib',
>'descriptive', 'simplex');
>+    public static $maximalibraries = array('stats', 'distrib',
>'descriptive', 'simplex', 'bitwise');
>
>Is needed for the base n work, or just another useful library of Maxima
>functions you'd like the option of having?!
>
>I've reviewed this, and it looks like a sensible inclusion, so I'm
>happy to have it here.  If the base n work needs it then we should just
>load it automatically, otherwise it will cause confusion if people
>don't have this, and then things break. 
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/maths/moodle-qtype_stack/pull/456#pullrequestreview-228918901

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

sangwinc:
Thank you very much indeed for these changes.

Matti and I spent a lot of time on the new parser, which we will need to really embed this work into STACK.  Finishing the new parser is my priority for this week.

PR #465, opened 2019-05-01
Instructions clarified and paths corrected

### LINKED ISSUE #464: Make it easier for teachers to install Maxima sandbox
The[ current explanation](https://github.com/maths/moodle-qtype_stack/blob/master/doc/en/CAS/STACK-Maxima_sandbox.md) for how to set up a Maxima sandbox is not easy to understand. There is the file [sandbox.wxm](https://github.com/maths/moodle-qtype_stack/blob/master/stack/maxima/sandbox.wxm) that could in principle make it easy, but it currently has hard-coded paths that assume the user has access to a moodle server. It would be good to have a version of this file that works well for users who do not.   Michael Kallweit (@m-r-k) suggested having different versions of the file for different operating systems to make it even easier for the user. Once these files are in place, the page explaining the setup can be drastically simplified.

m-r-k:
Improves the sandbox setup instructions. Fixes #464

### DIFF
--- a/doc/en/CAS/STACK-Maxima_sandbox.md
+++ b/doc/en/CAS/STACK-Maxima_sandbox.md
@@ -4,17 +4,21 @@ It is very useful when authoring questions to be able to test out Maxima code in
 That is to say, to run a desktop version of Maxima with the local settings and STACK specific functions loaded.  This is also used in [reporting](../Authoring/Reporting.md) and analysis of students' responses.
 To do this you will need to load your local settings, and also the libraries of Maxima functions specific to STACK.
 
-# STACK - Maxima sandbox (without access to a server)
+## Setup Maxima and wxMaxima
+
+Install Maxima (http://maxima.sourceforge.net/) and wxMaxima (https://wxmaxima-developers.github.io/wxmaxima/).
+
+## STACK - Maxima sandbox (without access to a server)
 
 If you don't have access to a STACK server then you will need to download the files.   Download all the STACK files from GitHub (git or as a .zip).   E.g. try `https://github.com/maths/moodle-qtype_stack/archive/master.zip`
 
-Unfortunately, there are a lot of .php files as well.  Most of the files you need are in
+The only files you need are in
 
     .../stack/maxima/
 
-Edit and use the file `.../stack/maxima/sandbox.wmx`
+In this directory open the file `sandbox.wmx` with wxMaxima and edit it to your needs.
 
-# STACK - Maxima sandbox (with access to a server)
+## STACK - Maxima sandbox (with access to a server)
 
 It is very useful when authoring questions to be able to test out Maxima code in the same environment in which STACK uses [Maxima](Maxima.md).
 That is to say, to run a desktop version of Maxima with the local settings and STACK specific functions loaded.  This is also used in [reporting](../Authoring/Reporting.md) and analysis of students' responses.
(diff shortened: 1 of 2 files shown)


doc/en/CAS/STACK-Maxima_sandbox.md
```diff
@@ -4,17 +4,21 @@ It is very useful when authoring questions to be able to test out Maxima code in
 That is to say, to run a desktop version of Maxima with the local settings and STACK specific functions loaded.  This is also used in [reporting](../Authoring/Reporting.md) and analysis of students' responses.
 To do this you will need to load your local settings, and also the libraries of Maxima functions specific to STACK.
 
-# STACK - Maxima sandbox (without access to a server)
+## Setup Maxima and wxMaxima
+
+Install Maxima (http://maxima.sourceforge.net/) and wxMaxima (https://wxmaxima-developers.github.io/wxmaxima/).
+
```
gustavdelius: On my system (Linux Mint) in addition to installing Maxima I also had to install the maxima-share package.

doc/en/CAS/STACK-Maxima_sandbox.md
```diff
@@ -4,17 +4,21 @@ It is very useful when authoring questions to be able to test out Maxima code in
...
 
     .../stack/maxima/
 
-Edit and use the file `.../stack/maxima/sandbox.wmx`
+In this directory open the file `sandbox.wmx` with wxMaxima and edit it to your needs.
 
-# STACK - Maxima sandbox (with access to a server)
+## STACK - Maxima sandbox (with access to a server)
 
 It is very useful when authoring questions to be able to test out Maxima code in the same environment in which STACK uses [Maxima](Maxima.md).
```
gustavdelius: There is duplication of text here. Also, do we still need all this complicated stuff? Isn't using sandbox.wmx always the easiest solution?

stack/maxima/sandbox.wxm
```diff
@@ -1,40 +1,43 @@
...
 STACK Sandbox
    [wxMaxima: title   end   ] */
 
+
 /* [wxMaxima: comment start ]
 This document loads the extra files needed for STACK.
 See https://github.com/maths/moodle-qtype_stack
 
-1. Download the files from github, either using git or as a zip file.
-2. Place the files somewhere they can be read, and edit the line velow to give the location.
```
gustavdelius: Is it ever necessary to change this? Does sandbox.wxm not automatically look for the files in its own directory?

doc/en/CAS/STACK-Maxima_sandbox.md
sangwinc: Gustav,
I think if we write parallel instructions on how to install maxima locally we will end up with outdated instructions which are wrong!  So, sorry this was a problem.

sangwinc:
We need to add the path to Windows, as wxMaxima on Windows does not always know its own path!

```
maximaplatform:"win"$
stacklocation:"c:/tmp/stack/stack"$
```

doc/en/CAS/STACK-Maxima_sandbox.md
gustavdelius: @sangwinc I think it is important to point out that just installing from the link provided is not enough, but that STACK needs additional maxima packages. I agree that we should not go into details of how to install those extra packages, we just should draw the reader's attention to the fact that they are needed.

PR #472, opened 2019-05-10
specify node on exponent of outer function

birdanja:
For the feedback "It looks like you have subtracted one off the power of the outer function instead of adding one to the power!" 
[My version] I would expect the student to have a false answer like this:
-5/12*(3*x-2)^-4
Here the student only had the wrong idea about the exponent, but then did everything right (within his logic).

[Original version] The original text seems to expect the student to have a false answer like this:
5*(3*x-2)^-4
Here the student would have made two mistakes. He had the wrong idea about the exponent and he forgot to consider the derivative of the inner function 3*x-2.

Am I missing something or do you agree, that the changes would make the prt more precise?

### DIFF
--- a/doc/en/Authoring/Authoring_quick_start.md
+++ b/doc/en/Authoring/Authoring_quick_start.md
@@ -304,7 +304,7 @@ If the first test is false, we will then perform the test in Node 2.
 Update the form so that Node 2 has
 
     SAns = diff(ans1, x)
-    TAns = 60*(3*x-2)^-5
+    TAns = 5*(3*x-2)^-5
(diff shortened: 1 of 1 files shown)


sangwinc:
Thank you very much indeed for pushing this suggestion.  Actually, I think we just didn't explain what we intended here very well.  So, I've just push a clarification to the docs.  I hope this makes sense?

sangwinc:
https://github.com/maths/moodle-qtype_stack/commit/d3b06a5d92415ab42f658927edde22804abe00ac

PR #477, opened 2019-05-14
RE: Clarify the author quick-start guide.

birdanja:
Thank you very much for the clarification and your fast response.

If I understood it correctly TAns still needs a factor 3 for the answer test to work correctly, so I added the differentiation of the inner function 3*x-2 to match with SAns = diff(ans1,x) in Node 2.


sangwinc:
Thanks!

PR #486, opened 2019-06-20
STACK: Inconsistent database defaults

mkassaei:
Hi Chris,
This issue was highlighted during a db-server refresh, the XMLDB "Check defaults" tool was run via /admin/tool/xmldb/ and the following inconsistencies were found in qtype_stack:

Table: qtype_stack_prts. Field: firstnodename, Expected '' Actual -

I have updated the corresponding upgrade code.
BTW, in upgrade code this was done in 2013 (version 2013030802), however, I did the patch before seeing this. And the issue does not appear anymore. In other words, somehow the upgrade on 2013 (version=2013030802) did not apply on our server.

Mahmoud

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -807,6 +807,19 @@ function xmldb_qtype_stack_upgrade($oldversion) {
         upgrade_plugin_savepoint(true, 2018120500, 'qtype', 'stack');
     }
 
+    $newversion = 2019061900;
+    if ($oldversion < $newversion) {
+        // Changing the default of field firstnodename on table qtype_stack_prts to drop it.
+        $table = new xmldb_table('qtype_stack_prts');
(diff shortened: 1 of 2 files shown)


timhunt:
Actually, we need to hold off on this for a bit. What Moodle does with columns like this (CHAR(_N_) NOT NULL DEFAULT _X_) is funny.

We found this problem using the built in check in Moodle (Admin -> Development -> XMLDB -> Check column defaults). That highlighted some discrepancies (which sometimes come from differences between code in install.xml and upgrade.php files, but in this case come form inconsistent behaviour in the Moodle DDL API.)

It is not a serious problem. It would just be nice to clean this up.

Sam Marshall is just doing some more investigation about what is going on here, so expect a revised patch when we have worked it out.

PR #495, opened 2019-07-19
Force math formulas input directionality to LTR, also in RTL mode.

nadavkav:
Suggested fix for RTL mode, in which all math formulas need to be input in LTR mode (text directionality)

### DIFF
--- a/stack/input/algebraic/algebraic.class.php
+++ b/stack/input/algebraic/algebraic.class.php
@@ -42,7 +42,7 @@ class stack_algebraic_input extends stack_input {
             'name'  => $fieldname,
             'id'    => $fieldname,
             'size'  => $this->parameters['boxWidth'] * 1.1,
-            'style' => 'width: '.$size.'em',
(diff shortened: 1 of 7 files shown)


sangwinc:
Thank you very much indeed.  Very sensible!

Should that be in the other input types as well.  What would you like to do about text areas for example?

nadavkav:
I was wondering if it might be better to add a class to the input element (and all other input types) and have this fixed with a CSS selector, and not inlined CSS, what do you think?

Also, we need to make sure that some of the question creator forms input fields are also keeping their alignment and directionality to the left in RTL mode.

nadavkav:
I have pushed a better fix for this issue, please review.

timhunt:
I think this is a good change.

The better fix is better ;-)

sangwinc:
Thank you,
Chris

nadavkav:
Thank you @timhunt and @sangwinc :-)

PR #515, opened 2019-11-06
Fix Behat fails caused by the MDL-66821 edit menu changes

timhunt:
Chris,

At the point where you start using Moodle 3.8, you will need these changes, or all the Behat tests will start to fail.

And, as soon as you are using 3.7.3 or 3.6.7 for development it will be safe to incorporate these changes, becuase the new steps were put into all Moodle branches with appropirate implementations.

The explantion is here: https://github.com/moodle/moodle/blob/e04a73ccc06e18d8d3b3661f8f9bc16911747830/question/type/upgrade.txt#L5. I am afraid it is all my fault, but a necessary evil.

### DIFF
--- a/tests/behat/create_preview_edit.feature
+++ b/tests/behat/create_preview_edit.feature
@@ -32,7 +32,7 @@ Feature: Create, preview, test, tidy and edit STACK questions
     Then I should see "Test STACK question"
 
     # Preview it.
-    When I click on "Preview" "link" in the "Test STACK question" "table_row"
+    When I choose "Preview" action for "Test STACK question" in the question bank
     And I switch to "questionpreview" window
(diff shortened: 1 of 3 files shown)


timhunt:
As an aside, now that the actions in the question bank are all in a menu, it becomes almost feasible to to add actions like "Question tests and deployed variants" and "Tidy question" to the menu for STACK questions. 

Probably, ... I was just about to write something hypothetical, but then I just went ahead and did https://tracker.moodle.org/browse/MDL-67153 and https://github.com/moodleou/moodle-qtype_pmatch/commit/2aefa8b5dcc7bab768f4707a4ffb7befcf4c2540. Should be easy to do the same in STACK when we have nothing better to do.

PR #534, opened 2019-12-18
STACK breaks if Moodle option 'forceclean' is turned on (e.g. View as Student)

### LINKED ISSUE #504: STACK breaks if Moodle option 'forceclean' is turned on (e.g. View as Student)
STACK breaks if Moodle option 'forceclean' is turned on (it's on the /admin/settings.php?section=experimentalsettings page).

This is basically because in the renderer, we call format_text too late. https://github.com/maths/moodle-qtype_stack/blob/50adb2c2b81422c77bc541049065299c5650395a/renderer.php#L80. We need to do that before replacing the PRTs and Inputs.

At some point Moodle is going to turn that option on by default (https://tracker.moodle.org/browse/MDL-62352), so we will have to fix this.

Also, if you use Moodle's Log in as feature, then that option gets forced on (https://github.com/moodle/moodle/blob/a672f021eaf0640c6f3b0094839c8c4fd3b2df7a/lib/moodlelib.php#L2798). So, if you Log in as another user, STACK is broken now. So, if anyone spots that, this is why.

HuongNV13:
This will resolve #504

### DIFF
--- a/renderer.php
+++ b/renderer.php
@@ -41,6 +41,27 @@ class qtype_stack_renderer extends qtype_renderer {
         // Replace inputs.
         $inputstovaldiate = array();
         $qaid = null;
+
+        // Get the list of placeholders before format_text.
+        $originalinputplaceholders = stack_utils::extract_placeholders($questiontext, 'input');
+        $originalfeedbackplaceholders = stack_utils::extract_placeholders($questiontext, 'feedback');
+
+        // Now format the questiontext.
+        $questiontext = $question->format_text(
+                stack_maths::process_display_castext($questiontext, $this),
+                $question->questiontextformat,
+                $qa, 'question', 'questiontext', $question->id);
+
+        // Get the list of placeholders after format_text.
+        $formatedinputplaceholders = stack_utils::extract_placeholders($questiontext, 'input');
+        $formatedfeedbackplaceholders = stack_utils::extract_placeholders($questiontext, 'feedback');
+
+        // We need to check that if the list has changed.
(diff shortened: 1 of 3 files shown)


renderer.php
```diff
@@ -33,6 +33,7 @@
 class qtype_stack_renderer extends qtype_renderer {
 
     public function formulation_and_controls(question_attempt $qa, question_display_options $options) {
+        $qtype = new qtype_stack();
```
timhunt: It is definitely wrong to manually create a new instance of the qtype class. 

The comment on get_input_names_from_question_text says "This method is needed for validation, and to construct the editing form, and looking at what it does (using data_submitted) means that it is not safe to use when a student is attempting the question.

Mind you, this function is already been called from initialise_question_instance, so the comment is wrong, and calling from initialise_question_instance is probably theoretically wrong too.

I think the logic we want, in both initialise_question_instance and here, is just what  stack_utils::extract_placeholders($questiontext, 'input') does. 

Do we also need to worry about feedback placeholders getting messed up?

Also, there really should be at least one unit test added (probably copy one test function in question/type/stack/tests/walkthrough_adaptive_test.php, and add make a version that turns on forceclean. Make sure it fails without this bug fix.

timhunt:
The latest version of the patch (commit 178945e4159881ca8123e090b4c2b9bc67130c1d) looks OK to me. Thanks Huong.

sangwinc:
Thank you very much indeed for this contribution Huong.

PR #541, opened 2020-01-14
Adding failable PHP 7.2, 7.3 and 7.4 unit test build targets

WillSkates:
Hello,

A small one solving:

1. PHP 7.2, 7.3 and 7.4 build targets for unit test runs. All of them are allowed failures.
2. [moodlerooms/moodle-plugin-ci is now abandoned](https://packagist.org/packages/moodlerooms/moodle-plugin-ci). Replaced with [blackboard-open-source/moodle-plugin-ci](https://packagist.org/packages/blackboard-open-source/moodle-plugin-ci) as recommended.
3. apt-get dependencies that aren't available in travis. Dependency list now matches [stack/4.3a](https://github.com/maths/moodle-qtype_stack/blob/stack4.3A/.travis.yml#L13).
4. Running Behat tests with moodle 3.6 as 3.7 has been released.

Should MOODLE_BRANCH=master be "MOODLE_BRANCH=MOODLE_37_STABLE" now that moodle 3.8 is out?

Kind Regards,
Will

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ addons:
   postgresql: "9.6"
   apt:
     packages:
-      - oracle-java8-installer
-      - oracle-java8-set-default
+      - openjdk-8-jre-headless
+      - chromium-chromedriver
       - maxima
       - maxima-share
       - texinfo
@@ -29,6 +29,18 @@ matrix:
 
     - php: 7.0
       env: TASK=PHPUNIT        MOODLE_BRANCH=MOODLE_32_STABLE   DB=pgsql
+    
+    - php: 7.1
+      env: TASK=PHPUNIT        MOODLE_BRANCH=MOODLE_32_STABLE   DB=pgsql
+
+    - php: 7.2
+      env: TASK=PHPUNIT        MOODLE_BRANCH=MOODLE_36_STABLE   DB=pgsql
+    
+    - php: 7.3
+      env: TASK=PHPUNIT        MOODLE_BRANCH=MOODLE_36_STABLE   DB=pgsql
+    
+    - php: 7.4
+      env: TASK=PHPUNIT        MOODLE_BRANCH=MOODLE_38_STABLE   DB=pgsql
 
     - php: 5.6
       env: TASK=PHPUNIT        MOODLE_BRANCH=MOODLE_33_STABLE   DB=mysqli
@@ -66,7 +78,7 @@ matrix:
 
     # Master here will need to be changed to MOODLE_36_STABLE once Moodle 3.7 has been released.
     - php: 7.1
-      env: TASK=BEHAT          MOODLE_BRANCH=master             DB=pgsql
+      env: TASK=BEHAT          MOODLE_BRANCH=MOODLE_36_STABLE   DB=pgsql
 
 
     - php: 7.1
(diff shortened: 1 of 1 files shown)


timhunt:
Active development is currently taking place on the stack43A branch, where this sort of travis config fix has already been done.

WillSkates:
@timhunt Thanks for replying Tim. I thought master was current as there are commits in master that are less than a week old. I'll go ahead and close this.

Are you still wanting a fix for: https://github.com/maths/moodle-qtype_stack/blob/master/.travis.yml#L22? The option to set a tcp port is missing from the installer. I could patch that for you upstream or in here.

Kind Regards,
Will

timhunt:
The current state of play is https://github.com/maths/moodle-qtype_stack/blob/stack4.3A/.travis.yml. Further improvements on that welcome.

We are very close to getting that green (https://travis-ci.org/maths/moodle-qtype_stack/jobs/636807387). The difficulty is that different versions of Maxima have subtly different behaviours in some situations, and the remaining 7 PHPunit fails are in that category.

WillSkates:
@timhunt sorry for taking a while to get back to you. Do you have a breakdown of the syntax changes between maxima versions. You could specify a specific version to test against in Travis rather than using the version supplied by apt. Could you please show me a version of Maxima that worked last?

timhunt:
It is more complex than just Maxima version. It appears to also behave differently depending on which Lisp interpreter was used as the runtime. Travis is in the Ubuntu/Debian world, which uses GCL for its Maxima package, but I don't think anyone uses that for real. The OU uses SBCL. Others use that or CLISP.

Apparently the unit tests currently all pass on Chris Sangwin's setup. @sangwinc, can you remind us what that is?

I think the only way to get that config here is to compile the Lisp and Maxima from source, but that should be scriptable.

If we could make that work, then acutally the ideal would be to configure Travis to run the tests against a wide sample of Lisps and Maximas, then we would really know the limitations of our support.

Thanks for looking at this.

aharjula:
For those that care, Aalto uses SBCL in new installs but still has some environments with CLISP. Primarily due to the ease of automating image-generation in the former.

We do not typically compile the LISP from scratch, the Debian package of SBCL works just fine and one can compile Maxima for it quite quickly (well depends on how large a `-j`-switch one can use in the travis machine).

WillSkates:
@timhunt Can you check the installed lisp interpreter as you install the plugin and display a notice if the plugin is already installed and the lisp interpreter isn't supported?

Do you know which LISP interpreter @sangwinc is using.

@aharjula It may well be that in order to work consistently, the plugin will have to detect which interpreter is being used and then parse output from Maxima into a consistent format. Unless of course your questions are dependent on the output you get with SBCL. In which case the tests need to be expanded upon to include the LISP interpreter in their matrix.

sangwinc:
In Edinburgh we are using Maxima 5.38.1 using Lisp GNU Common Lisp (GCL) GCL 2.6.12.
Chris

WillSkates:
Build config: https://github.com/WillSkates/maxima-travis/blob/master/.travis.yml
Travis Example: https://travis-ci.org/WillSkates/maxima-travis/builds/648538738

timhunt:
That looks promising. Thanks! Now to integrate it with STACK's travis.yml file on the dev branch. I will probably be too busy to try for a few days.

WillSkates:
@timhunt No problem. Apologies as I've got other commitments that prevent me from having a go at the work myself.

PR #544, opened 2020-01-24
Quote HTML entities in alt attribute.

perhov:
When plotting an expression containing greater than and less than symbols, the generated HTML output is malformed, and part of the alt text of the img tag is displayed below the plot in the browser.


aharjula:
Thank you, seems good to me. Lets just ask if Chris has something to say.

Chris? Was there any additional logic related to this elsewhere or maybe existing issues and workarounds to check? I have a feeling that something related has drifted past my eyes.

sangwinc:
Thank you very much for this fix.  I can't think of anything related.

PR #585, opened 2020-06-04
Fix Question Variables help link

aspark21:


sangwinc:
Many thanks indeed for spotting that.

Issue #468, opened 2019-05-02
New input type: data entry/resizable matrix


sangwinc:
We want to better support colleagues in physics who would like to accept data from students in a lab situation.

Based on some ideas we had for resizable matrices we would like the following features.

Validation options
1. require min/max number of rows/columns (If they are equal, that forces, e.g. the number of columns)
2. require min/max accuracy dp/sf by column/row.  That is, require column 1 to be to 2 sfs, but column 2 to 3sf.  This validates data entry.
3. As in 2 but require units, with various options such as compatible units.

The result would be a matrix.

We would also need better answer tests for dealing with sig figs in matrix entries.

sangwinc:
We should also implement issue #469 and make sure this is also available here.

dhilsen:
The ability to enter physics lab data in Stack would be very useful. This could be valuable for both a physical lab involving apparatus or a computer simulated lab. 

I would envision the instructor would be able to specify the number of columns and rows, along with the physical quantities that define the columns. For example, a Hooke's law lab would have the instructor establish two columns, one for the weight applied (to vertically oriented spring), and a second for the measured extension of the spring. Specifying 10 rows would require the student to gather 10 data points. It would be important to have Stack validate entries in each column for the correct number of significant figures and correct unit. 

With an investigative lab the goal is often to have the student find the relationship which exists between the involved quantities. I think most would agree this is more valuable than simply verifying an already known relationship. After the data has been gathered perhaps Stack could generate a plot of the data. The student would then enter a mathematical model and Stack would evaluate if the data fits the model. If the data does not fit the model the student would be required to enter a new model. This would of course require that some specified tolerance be taken into consideration by Stack to allow for error in measurement.

christianp:
I've had such headaches about the right design for a customisable matrix input in Numbas.
There's a data grid (spreadsheet) component called [handsontable](https://handsontable.com/) that I've been meaning to have a go at using for a long time. That would work well for entering lab data.

PR #632, opened 2020-09-16
Fix for JavaScript errors in Deferred feedback mode

timhunt:
I think this is the fix for the JavaScript error that George reported by email. However, I have not really tested it enough yet (and I am not sure I have time to) so not adding it to master yet.

I note that Behat was failing on Travis, which might have been because if this issue.


timhunt:
Travis build here: https://travis-ci.org/github/maths/moodle-qtype_stack/builds/727833805

I had to move grunt check to CODEKNOWNFAILS becuase Chris has copied and pasted all the jQuery-UI CSS into STACK's CSS file. I am not sure that is a good idea.

sangwinc:
Thanks Tim, for looking at this so quickly.

> Chris has copied and pasted all the jQuery-UI CSS into STACK's CSS file.
Sorry if I messed up here!  Do you mean this?
https://github.com/maths/moodle-qtype_stack/blob/master/styles.css#L516
Looking at the history, this was to fix issue #468.

aharjula:
Have to wonder what we would do with jQuery-UI CSS if we do not have the matching JS (and we do not want it) and based on my trials that will not mix well with Moodle at all. Maybe someone should track the uses of those style-classes and rewrite them and only them with a STACK-specific prefix in the name.

One would guess that that particular injection of CSS only meant to inject [these](https://github.com/maths/moodle-qtype_stack/blob/0cdabea7cdce84842d0924d0cc1c590dfcaed521/styles.css#L815-L865) but even then where are those used?

Chris do we have a style-test question that has all the possible STACK-styles in it? All the inputs with all the style affecting options and all the feedback-forms as well as all the new system-level styles? Using such it should be simple to go through the document and spot the rules in that CSS that are not being used, also would work as a style document for a quick check of browser compatibility. As you have been adding style maybe you should produce such a question and take screenshots of it being rendered in the form you expect it to look. Screenshots for the Moodle themes that come with the default install should be enough.

sangwinc:
Tim,
I think we can now delete this branch?  Thank you so much for fixing this so quickly.
Chris

timhunt:
Yes. Branch can now go. In fact, I just clicked the button.

PR #633, opened 2020-09-19
Fix to 0482d52 - Include styles.css on docs.stack-assessment.org

malthefogsporring:
This fixes the issue raised on commit 0482d52, to include STACKs stylesheet on the online docs. The issue was that raw.githubusercontent serves up styles.css as a txt file. Here I use combinatronics instead and it works.


sangwinc:
Many thanks indeed for looking at this.
Chris

Issue #674, opened 2021-01-08
Strange error involving ssubst


NeilStrickland:
Has anyone else seen this?
``` CASText failed validation. (%i5) STACK-OUTPUT-BEGINS> `ssubst': unsuitable start or end position. -- an error. (%i6) (%o6) ```
This is extremely strange, for a number of reasons.
1. When submitting an answer to a certain large and complex question, I get a `TIMEDOUT` error.  (This is fast, even after clearing the cache, so it does not seem to be a real timeout.)
2. Running the same code in the CAS Chat  window gives an error as above
3. If I run Maxima from the command line, load `stackmaxima.mac` and run the same code, I get no error, either on my Linux server or my Windows desktop.  On the server I am using the same Maxima executable as is invoked by Stack via MaximaPool.
4. I can make the error go away by reducing the size and complexity of the question.  No specific piece seems to be critical, but it is not a straightforward count of characters either.  Some size-reductions work and others do not, with no obvious pattern.
5. The function `ssubst` can be called with three or six arguments.  The error message only makes any sense for the six-argument form.  I do not directly call `ssubst`.  The Stack source code has a number of calls to the three-argument form, which all seem completely harmless, and no calls to the six-argument form.  
6. In the Maxima source code, `ssubst` is defined in `share/stringproc/stringproc.lisp`.  There are a couple of functions in the same file that call `ssubst`, but none of them seem to be relevant for this issue.  As far as I can tell, there are no other calls to `ssubst` anywhere else in the Maxima source code.

Unfortunately, it is hard for me to post a self-contained question file as I am using a modifiied version of Stack.  However, I may try to find a way around that if no one can offer other suggestions.

NeilStrickland:
This seems to be a bug in Maxima which I have reported at https://sourceforge.net/p/maxima/bugs/3700/.  It should be possible to work around it by refactoring the calls to `ssubst` in `stackmaxima.mac`.  I will sort out details of that shortly.

aharjula:
Thank you for tracking this down. Do you happen to know if it is just the GCL version that has this problem? For me trying in SBCL (5.42.2) even inputting that string is an interesting thing, it seems to parse it for quite a long time, actually does not even complete (in half an hour). I have to wonder though if it is the shell that has trouble parsing that and if that string would work if being generated programmaticaly from smaller bits. Quick split to small fragments for `sconcat` did not do the trick though.

[...]

PR #676, opened 2021-01-10
Refactor calls to ssubst

NeilStrickland:
This fixes https://github.com/maths/moodle-qtype_stack/issues/674.  The function `ssubst(a,b,s)` generates an error if there are too many instances of `b` in `s`; this is a bug in Maxima (in some versions on some platforms, at least).  This comes into play if we have complex questions that generate generate LaTeX strings with many backslashes, for example.  So we change the functions `stackjson_stringify`, `str_to_html` and `str_to_js` to use `charlist` and `simplode` rather than `ssubst`.

### DIFF
--- a/stack/maxima/stackstrings.mac
+++ b/stack/maxima/stackstrings.mac
@@ -197,22 +197,17 @@ stack_string_parse_number(somestring) := block([c, b, phase, neg,nege, tmp, i],
  return(c)
 )$
 
-
 /* Takes pretty much anything and turns it to a JSON string */
-stackjson_stringify(obj) := block([tmp,r,l],
+stackjson_stringify(obj) := block([tmp,r,l,f],
  r:und,
+ f(c) := if c = "\\" then "\\\\" else if c = "\"" then "\\\"" else if c = ascii(8) then "\\b" 
+    else if c = ascii(9) then "\\t" else if c = ascii(10) then "\\n" else if c = ascii(12) then "\\f" 
+    else if c = ascii(13) then "\\r" else c,
  if is(obj=und) then r:"null"
  else if is(obj=false) then r:"false"
  else if is(obj=true) then r:"true"
  else if stringp(obj) then (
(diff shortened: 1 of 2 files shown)


aharjula:
Seems good to me, should work with unicode as well. Probably even faster. We have to have some way for dealing with multiple character long substitutions in the future, hopefully the Maxima side ticket leads to that being fixed.

Should be merged to dev though.

NeilStrickland:
I see that there are build errors, which do not seem to have any relationship with the things that I changed.  I have not understood the Travis framework at all yet.  If anyone can explain briefly how I should think about these errors, that would be appreciated.

aharjula:
I hope Chris or Tim can explain Travis at some point. In any case, I looked at this again and started to worry about this redefining the function `f`. I have to wonder if that leads to trouble in questions that include that function or expect students to define things with it.

NeilStrickland:
> I hope Chris or Tim can explain Travis at some point. In any case, I looked at this again and started to worry about this redefining the function `f`. I have to wonder if that leads to trouble in questions that include that function or expect students to define things with it.

Indeed, after some investigation I find that Maxima's scoping rules are a bit strange and different from what I thought, so we should refactor around that.  I will deal with that at some point.

timhunt:
First, the issues are not wit Travis. All travis is doing is running the automated tests with different versions of Moodle.

The Behat fail is probably spurious. I think Chris renamed the demo course, whithout updating the test that uses it.

The PHPunit failure with just one moodle version is becuase they updated the PHPunit framework there, and so the test need small update.

Anyway, we probably need to move away from Travis-CI, since they are no longer as friendly to open source as they used to be. (They got bought out.) However, people in the Moodle world have worked out how to swtch to 'GitHub actions' instead. I made #677 to do that. I am busy this week, but I can probably do it some time next week.

sangwinc:
Matti, you were spot on about problems with redefining `f` in the comment made above.

PR #684, opened 2021-02-07
Fix table in Potential_response_trees.md

kasmar00:
There was a missing `|` in table, making the rendering of table fail.


sangwinc:
Thanks, this is much appreciated.

Issue #713, opened 2021-06-05
Documentation about SRegExp outdated (?)


sm8ps:
I have been playing with the SRegExp test -- successfully in the end. I recall discussions about the transition from (PHP-?)RegExp and it seems that the documentation is a bit behind the current state.
[It reads](https://github.com/maths/moodle-qtype_stack/blob/master/doc/en/Authoring/Answer_tests.md#sregexp): "Currently this is not provided as a separate answer test so you will need to use this predicate in the question variables and check the result against the expected value, or supply the predicate as an argument to an answer test." -- This remark made me jump through quite some hoops in order to circumvent a problem that does not seem to exist anymore in my version 4.3.8 of Stack where the SRegExp test is available and works well. I do not have any good suggestion about how to describe the current state though.

sangwinc:
Very sorry about the incorrect docs here. Thank you very much indeed for letting me know.
Chris

sm8ps:
Thank you for providing this great tool!

shinuito:
> Very sorry about the incorrect docs here. Thank you very much indeed for letting me know.
> Chris

Did the recent update break something on the documentation page? It seems like the top bar with the search is broken for me now and squashed into the top left corner. There's also a console error ```Uncaught ReferenceError: Missing element: expected "#__config" to be present```

malthefogsporring:
Thanks for pointing this out @shinuito. The above pull request fixes this issue.

PR #714, opened 2021-06-12
Updates custom overrides of mkdocs-material files

malthefogsporring:
This fixes the style issue with the online docs identified by @shinuito in maths/moodle-qtype_stack#713. One of the updates to mkdocs-material had, among other things, changed the names of some of its css styles. Because our implementation includes custom overrides of some of the material files, the website would not work until we referred to the new style names locally.

### DIFF
--- a/.github/workflows/deploy_website.yml
+++ b/.github/workflows/deploy_website.yml
@@ -13,6 +13,6 @@ jobs:
           python-version: 3.x
       - run: pip install mkdocs-material
       - run: pip install https://github.com/mitya57/python-markdown-math/archive/master.zip
-      - run: pip install pip install mkdocs-exclude
+      - run: pip install mkdocs-exclude
       - run: pip install mkdocs-redirects
       - run: mkdocs gh-deploy --force
--- a/doc/en/Developer/Website.md
+++ b/doc/en/Developer/Website.md
@@ -53,7 +53,7 @@ You can also edit the theme directly. Any file in the `site_overrides` folder wi
 Before adding major style changes to the online docs, you are encouraged to test your changes locally. For this, you will need to install MkDocs and all the required extensions.
 
 1. [Install MkDocs](https://www.mkdocs.org/), including its requirements.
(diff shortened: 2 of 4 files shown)


malthefogsporring:
We will have this problem again the next time mkdocs-material makes any style changes that conflict with our custom file overrides, so I would be keen to find a permanent solution. Our webhook, which rebuilds the website anytime master is pushed to, installs mkdocs-material prior to building, so if possible we could change it to always install the current (12/6/21) version of mkdocs-material? The tradeoff would be that we would miss out on any bug fixes from the mkdocs-material team. Thoughts?

sangwinc:
Malthe,
Thank you very much indeed for fixing this.  This is the classic problem of dependencies which change!  Really, the only general solution is to keep on top of changes to the packages on which we rely.  Or, remove the dependency on those packages!

Many thanks,
Chris

shinuito:
Hi,

Thanks for getting back about this. 

I can see that the commit was merged yesterday. Is it something that takes a bit of time to rebuild? I can see the styling looks a bit better now but the search bar is still missing. I can see it's now the case that the search icon (magnifying glass) and input box are both hidden with display:none and even unhiding them it doesn't seem to function. There's also still this console error in Chrome/firefox saying "Uncaught ReferenceError: Missing element: expected "#__config" to be present" which maybe is stopping the search from working?

Thank you for getting onto to fixing this though. It was a bit of work learning to use STACK through the documentation on github. It was quite tricky to find things until I realised I could search within the repo itself, but even then one still has to find the right .md files in the search. 

The new site with its documentation and search is of course much easier to navigate and find things :)

All the best,

Stephen

sangwinc:
Also, another user just emailed me because they just noticed the search function appears to have vanished, and the link to the quick start guide from https://stack-assessment.org/GetStarted/ doesn’t appear to work any longer.

malthefogsporring:
Thanks both, I've added another pull request that fixes these two issues - #719

PR #719, opened 2021-06-15
Fix search and linking issues on online docs

malthefogsporring:
This fixes issues with the online docs where the search bar was invisible and where directory urls were not being used, leading to broken links.


sangwinc:
Thanks again Malthe for stepping up and contributing to this so quickly.

Issue #720, opened 2021-06-24
6 failing unit tests with 2020120600 stack version


golenkovm:
Hi guys,

Has anyone seen these failing unit tests before?

```
Moodle 3.9.7+ (Build: 20210520), 49d363745316c3eb77f003ba086d33ea86fc0347
Php: 7.4.3, pgsql: 11.5 (Debian 11.5-1.pgdg90+1), OS: Linux 5.4.0-72-generic x86_64
PHPUnit 7.5.20 by Sebastian Bergmann and contributors.

.............................................................   61 / 3352 (  1%)
.............................................................  122 / 3352 (  3%)
.............................................................  183 / 3352 (  5%)
.............................................................  244 / 3352 (  7%)
.............................................................  305 / 3352 (  9%)
.............................................................  366 / 3352 ( 10%)
.............................................................  427 / 3352 ( 12%)
.............................................................  488 / 3352 ( 14%)
.............................................................  549 / 3352 ( 16%)
.............................................................  610 / 3352 ( 18%)
.............................................................  671 / 3352 ( 20%)
.............................................................  732 / 3352 ( 21%)
.............................................................  793 / 3352 ( 23%)
.............................................................  854 / 3352 ( 25%)
.............................................................  915 / 3352 ( 27%)
.............................................................  976 / 3352 ( 29%)
............................................................. 1037 / 3352 ( 30%)
............................................................. 1098 / 3352 ( 32%)
............................................................. 1159 / 3352 ( 34%)
............................................................. 1220 / 3352 ( 36%)
............................................................. 1281 / 3352 ( 38%)
............................................................. 1342 / 3352 ( 40%)
............................................................. 1403 / 3352 ( 41%)
............................................................. 1464 / 3352 ( 43%)
............................................................. 1525 / 3352 ( 45%)
............................................................. 1586 / 3352 ( 47%)
............................................................. 1647 / 3352 ( 49%)
............................................................. 1708 / 3352 ( 50%)
............................................................. 1769 / 3352 ( 52%)
............................................................. 1830 / 3352 ( 54%)
............................................................. 1891 / 3352 ( 56%)
............................................................. 1952 / 3352 ( 58%)
............................................................. 2013 / 3352 ( 60%)
............................................................. 2074 / 3352 ( 61%)
.....................F....................................... 2135 / 3352 ( 63%)
....................................F........................ 2196 / 3352 ( 65%)
............................................................. 2257 / 3352 ( 67%)
[...]

PR #721, opened 2021-06-25
Fix unit tests for Maxima 5.41.0 and plugin version 4.3.8

golenkovm:
This is a fix candidate for the issue #720


aharjula:
Sorry but no `3.0E+2` -> `3.E+2` is not something we want that is a sign of bad float handling in a LISP+Maxima combination. The test is not wrong the logic is if it cannot provide the correct output for that LISP+Maxima combination.

sangwinc:
Matti,
Sorry, but my WIP commit just added in ignoring the missing zero, e.g. here: https://github.com/maths/moodle-qtype_stack/commit/7d039d9ef0e58d0f7cf034f52184d3414ab5c55f#diff-cbec917988f4e4d83d380ce16ce24839e5eafe01528dbfabf0455e1883157bdeR202
If you think I've made a mistake here, please let me know and I'll revert this WIP.

aharjula:
Chris,
Well that fixes the tests, but should we not also fix the presentation for actual use? Is this not a case that only applies to the formatting rules that do not actually define the number of digits to display `"~a"` etc., maybe the Maxima side logic we have for formating the numbers should include special handling to fix those patterns if the problem is present in the LISP. Maybe the Maxima side logic sould simply do a test format of a suitable float to detect if this problem exists and if it exists replace the function with one that overrides those specific patterns.

cameron1729:
Since #935 has been merged, this one can be closed :)

PR #734, opened 2021-08-30
add an example of "potentially redefined in unclear substitutions"

georgekinnear:
This adds a short example to the docs, to explain one way to adapt code to suit the new validator.


sangwinc:
Thanks George, much appreciated.
Chris

PR #756, opened 2021-11-09
Set a time limit on CAS connections when using MaximaPool

timhunt:
I just noticed from our monitoring, than when the MaximaPool server cannot be reached, it just sits there for a long time, much longer than the CAS timeout set in the Admin settings. So, adding this timeout probably helps things work more smoothly when MaximaPool is not working smoothly.


timhunt:
In the end, we never used this, so probably best not to merge it!

PR #768, opened 2022-01-17
Add NO_OUTPUT_BUFFERING to two more scripts

timhunt:
I think these two scripts would also benefit form progressive output.


timhunt:
Also, a second commit to fix a couple fo failing Behat tests.

sangwinc:
Thanks Tim

Issue #770, opened 2022-02-01
Alt-text checking gives false-positives on STACK plots


timhunt:
At least on our system, the HTML for STACK plots is

```html
<img src='!ploturl!stackplot-3852702702-79504506.svg' alt='The straight line with gradient -1/3 passing though the points (5,-5)' width='450' />
```

But the regex used is in the alt-text checking is

```php
$r2 = "/(alt|title|src)\s*=\s*(\"[^\"]*\")/i";
```

So you get false-positive warnings because ' does not match ".

I'll do a pull request.

sangwinc:
This has all been merged into the main codebase.  Thanks @timhunt .

PR #771, opened 2022-02-01
Fix alt-text checking #770

timhunt:
Hi Chris, this is the fix for the issue I reported this morning.

We are in a period of testing and bug-fixing at the OU right now, so if you are able to hit the  merge button relatively soon, that would help us. Thanks.


sangwinc:
Thanks Tim.

PR #781, opened 2022-03-30
blockquote CSS rule must be scoped to only affect STACK

timhunt:
I also dealt with some errors and warnings the PHPStorm spotted.

### DIFF
--- a/styles.css
+++ b/styles.css
@@ -116,10 +116,7 @@ body.pagelayout-embedded .que.stack .formulation .questiontestslink {
     border: 0;
     border-left: 2px solid #000;
     border-right: 2px solid #000;
-    border-top-left-radius: 1em;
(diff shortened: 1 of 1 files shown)


timhunt:
Hi Chris. This generic style rule in STACK broke styling of content in other parts of our Moodle site. We need to make this rule more specific.

And, strictly speaking, we should go through all the rules in the CSS file, and ensure they cannot mess up other things. However, at least most of the other rules like .proof-step have a lot probability of conflicting.

But, strickly, every style rule in a plugin should start with `.que.stack`, or `body.path-question-type-stack...` or `body#page-question-type-stack-...`.

sangwinc:
Thanks Tim.

timhunt:
Thanks for the speedy review and merge Chris.

PR #786, opened 2022-04-11
Localization from Decimal / List Separators

Aron4u:
Hey Moodle-Stack,

I wrote the following prototype for the localization of decimal separators. 
In german you write 1,5  instead of 1.5. And this can be confusing for students.

The basic Idea is that you convert on the frontend UI from 1.5 to 1,5. And for lists you convert from [1.5,2,3] to [1,5;2;3]. 

If the the user chooses the german language, js creates a new Element that the input will be inserted from the user. The old input field is hidden. Any input from the server into this field is converted to the new Element. Any changes done to the new Element are inserted into the original input field. 

The next problem was finding a good slot where maxjax is 'done' so that elements exist etc. But the actual implementation is now quite some time ago so I don't remember the details here to much. 

My colleague (a teacher) tested it a bit an found no issue. 

Finally:
I currently do not have to much time on my hands, so I can't promise fast updates/answers here. Not sure if I can spend time fixing / coding. 

commit 76068b38941e008e8733e32b48db942b728dbe6e
Date:   Fri Jun 4 21:16:50 2021 +0200

    fix shadow element having _translate id
    move copy before new assignment

commit aae7668f190868396d8a5c1562e3baba231475bc

Date:   Sat May 1 19:07:57 2021 +0200

    fix issues related to validation fields
    rework script processing to use use Reprocess
    at the end of loading and modifing each script math content

commit 7dea705483ccef72ff2bb1cff408bc9e74e27793
Date:   Sun Apr 11 17:18:27 2021 +0200

    workaround fix for firefox

commit c2ac27d745d2119730c7c380f3e6ee59b3da402b
Date:   Sat Apr 3 16:17:07 2021 +0200

    Mostly Javascript Solution to localization,
    working for german / de
    working prototype


sangwinc:
We've solved this issue in a rather different way as issue #789.

PR #800, opened 2022-05-24
Update Maxima.md

anst-i:
Removed duplicate entry in list of Maxima commands defined by STACK: `exdowncase` is already described in line 93.


sangwinc:
Many thanks!

PR #812, opened 2022-07-05
Option to skip trying to build maxima-optimised when installing for test

timhunt:
We were getting a problem with fatal errors on our CI server when trying to install Moodle including the STACK code on a system without Maxima.

This provides a way for people to avoid those fatal errors. It would help us if you are prepared to merge this. Thanks.


sangwinc:
Very good suggestion Tim,
Thanks.

PR #828, opened 2022-08-04
Add support for Moodle version < 400

nadavkav:
Added include vle_specific.php, as it was needed for stack_determine_moodle_version()
Added support to Moodle version < 400, when calling question_preview_url()


sangwinc:
Thank you so much for sorting this out. I really appreciate it. I was about to sit down this morning and do this myself.
Chris

PR #834, opened 2022-08-08
change param-order and return type

nisaDoc:
Hello,
As of PHP 8.0, a deprecation log item is thrown when a required parameter follows an optional parameter.
By changing the parameter order, that can fix this problem.
Besides, the return type of php_string_to_maxima_string and maxima_string_to_php_string have a problem: "Expected type null|string. Found a"
So, I have also changed its return type into null|string.
Please have a look of these suggested solutions, or feel free if you any better ideas to fix the problems.

Thx and cheers,
Nisa

### DIFF
--- a/stack/cas/castext2/block.interface.php
+++ b/stack/cas/castext2/block.interface.php
@@ -96,7 +96,7 @@ abstract class stack_cas_castext2_block {
      * list of PRTs. What it contains in use case specific. One can typically assume
      * that it contaisn the same values as the simillar aray for `compile`.
      */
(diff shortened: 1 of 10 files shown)


sangwinc:
Nisa,
Thank you so much for making the effort to send us a pull request, much appreciated.
Chris

nisaDoc:
oh my bad. yes I have seen your solution for these problems. I think my pull request is no more necessary.
Thks

timhunt:
Closing, based on the last comment.

PR #838, opened 2022-08-18
Iss824

aharjula:
Unable to come up with more fixes that would not have more significant side effects, so this is what we should have as it restores much of the expected behaviour of `ev`. The use of `ev(...,eval)` would restore the rest, but as that can only be applied when it is safe to apply, and it is practically impossible to find out when it is safe we cannot apply it automatically.


sangwinc:
Thanks Matti,
I'm actually away on holiday at the moment with minimal internet access, so I didn't see this PR.  Thanks!  I agree.
Chris

PR #851, opened 2022-09-07
Avoid reusing h1

georgekinnear:
Having more than one h1 on the page breaks the "table of contents" in the online docs https://docs.stack-assessment.org/en/Authoring/Answer_tests/


sangwinc:
Thanks George.

Issue #849, opened 2022-09-05
Unable to use simp:false within CASText


georgekinnear:
I would often use something like `{@(simp:false, 1+1)@}` within a worked solution to show the details of a step (e.g. the question in #848). That seems not to work any more - if simplification is on at the end of the question variables, the output appears as "2".

So, it seems that it is no longer possible to turn off `simp` within subsequent CASText.

## Example

Question variables:
```
var1:1+1;
simp:false;
var2:1+1;
```
Question text | result with `simp:true` at end of question vars | result with `simp:false` at end of question vars
------|--------|----------
{@var1@} | 2 | 2
{@var2@} | 2 | 1+1
{@1+1@}| 2 |  1+1
{@(simp:false, 1+1)@} | 2 |  1+1
{@(simp:true, 1+1)@} | 2 |  2

I'm concerned that I will now have lots of questions with unhelpful worked solutions, because the steps that I'd tried to show are now being simplified. Plus, this change in behaviour is not picked up by bulk question tests, so it will take a bit of detective work to find them all.

## Workaround

To get selective control of simplification in the question text / worked solution, it seems that you must now have `simp:false;` at the end of the question variables.

aharjula:
How unfortunate. this is the primary reason why that no longer works, the other reason is that solving #844 requires us to detect the simp value from the expression, and we can only reliably detect the value from the end of the expression (we could catch it during runtime... however I would rather not for various reasons):
```
Maxima restarted.
(%i7) simp;

(%o7) true
(%i8) (simp:false,1+1);

(%o8) 1+1
(%i9) simp;

(%o9) false
(%i10) 1+1,simp:true;

(%o10) 2
(%i11) simp;

(%o11) false
(%i12) 1+1;

(%o12) 1+1
```
Note how the simp value outside the evaluation of `%i8` changed i.e. `%i9` now returns `false`. Whereas the way one can tune the simplification in `%i10` does not affect the global simp. 

Basically, if you need to tune evaluation details in CASText injections, you should use evaluation flags and write `{@1+1,simp=false@}` that will make sure that the modification of simplification only applies to that single injections evaluation and does not leak out nor gets wrongly detected.

[...]

PR #854, opened 2022-09-07
add more details about selective simplification in CASText #849

georgekinnear:
I expanded on the details using suggestions from https://github.com/maths/moodle-qtype_stack/issues/849

@aharjula - please correct me if I've said anything wrong!

### DIFF
--- a/doc/en/CAS/Simplification.md
+++ b/doc/en/CAS/Simplification.md
@@ -52,17 +52,45 @@ This includes the answer tests as needed.
 The level of simplification performed by Maxima can be controlled by changing Maxima's global variable `simp`, e.g.
 
     simp:true
-
-This variable can be set at the question level using the [options](../Authoring/Options.md) or for each [Potential response tree](../Authoring/Potential_response_trees.md). Within [CASText](../Authoring/CASText.md), the setting can be controlled by appending `simp=...`, for example: 
-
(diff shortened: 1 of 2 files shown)


doc/en/CAS/Simplification.md
```diff
@@ -52,17 +52,45 @@ This includes the answer tests as needed.
...
+    
+will give \(a=2\).
+
+### Within CASText (question text, general feedback, etc.)
+
+Sometimes it is useful to control the level of simplification applied to expressions included within [CASText](../Authoring/CASText.md) using `{@...@}`.
+In particular, to show steps in working, it is often necessary to turn simplification off.
+
+Note that it is not possible to reliably control the value of `simp` within the CAS expression itself: for instance, `{@(simp:false,1+1)@}` will display as \(2\) if `simp` is set to `true` at the question level.
+(This is because of the way STACK assembles all the CASText code into a single Maxima call.)
```
aharjula, line 74: Hopefully, this will be fixed for this simpler case for 4.4.2. At this moment, we are having a problem detecting the simplification setting in use in this form of expression, but we'll see if we can fix that.

While we do indeed turn the whole thing into a single expression, we can control simplification inside that expression we just need to detect when we need to do so.

sangwinc:
Thanks George.

PR #872, opened 2022-10-31
Fix theme issue for Question tests & deployed variants

AnupamaSarjoshi:
Hi Chris,

I have made a small change to fix the theme issue where there was a mismatch between number of table header columns and the number of table data columns in 'Question tools and deployed variants' tool.

Could you please review?

Thanks,
Anupama


sangwinc:
Many thanks indeed!

Issue #880, opened 2022-11-15
With Moodle 4.0, duplicating a STACK question loses the deployed variants and question tests


timhunt:
To reproduce:

1. Make a STACK question.
2. Add some deployed variants and question tests.
3. In the question bank, Duplicate the question.
4. Look at the question tests and deployed variants for the copy.

Expected result: they have been copied.

Acutal result: they are not there.

In the CI config, we have

      - name: Behat features
        continue-on-error: true # Currently fails. We really ought to get this passing.

so, we were just ignoring the fact we had a Behat test which showed this was broken. (Scenario: Test duplicating a STACK question keeps the deployed variants and question tests)

timhunt:
My fix was merged some time ago, so closing this now.

PR #881, opened 2022-11-15
Fix saving the deployed variants and question tests when a question is duplicated #880

timhunt:
Also, this gets all the Behat tests passing, and puts them back into the CI build properly. The other one with issues was the one that tested restore.


timhunt:
I missed the comment about mariadb being so much slower to run CI, so I just updated things to put must runs back to Postgres.

PR #883, opened 2022-11-16
Fix questiontestreport DB query to fetch all attempts of all versions

AnupamaSarjoshi:
Hi Chris,

The DB query in questiontestreport for Moodle 4.0 was fetching all the question attempts irrespective of the question id. Because of which, we were getting the below warning messages in the 'Basic question use report'. Have made the changes to fetch the question attempts considering all the question versions.

Could you please review the same?

Basic question use report:

Warning: Undefined array key 11 in <path>\question\type\stack\question.php on line 312
Warning: Undefined array key 10 in ...\question\type\stack\question.php on line 312

Thanks,
Anupama

### DIFF
--- a/questiontestreport.php
+++ b/questiontestreport.php
@@ -125,52 +125,37 @@ flush();
 // Later we only display inputs relevant to a particular PTR, so we sort out prt input requirements here.
 $inputsbyprt = $question->get_cached('required');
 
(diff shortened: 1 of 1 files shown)


sangwinc:
Many thanks indeed, that is really helpful.

PR #895, opened 2022-12-12
Update Authoring_quick_start_4.md

timlowe:
Correct latex code: {...} around exponent. (Mainly as a submission test).
Note: GitHub preview does not seem to run MathJax, so I can only assume it will work!


sangwinc:
Thanks Tim,

PR #925, opened 2023-02-22
Fix cartesian_product not working with maxima >=5.43

### LINKED ISSUE #924: `cartesian_product` function broken in maxima 5.43 and after
In maxima commit `c02a5b182`, the maxima function `cartesian_product_list` is introduced. To avoid code duplication with the existing `cartesian_product` function, a `cartesian-product` lisp function was written.

When it was introduced, the lisp function of the name was removed from the `logic` package, since the new `cartesian-product` function seems to be able to deal both with maxima lists and lisp lists. However, stackmaxima still contains an old `cartesian-product` implementation in `stack_logic.lisp` that does not work on maxima lists and therefore breaks the `cartesian_product` maxima function if used with maxima version >5.43.

Maybe the definition of `cartesian-product` in `stack_logic.lisp` could be guarded by an `(unless (fboundp 'cartesian-product) ..)` or checking the maxima version through `*autoconf-version*`?

Lennart-Kramer:
With maxima release 5.43, cartesian-product was added within maxima and was broken because stack_logic.lisp defined its own incompatible version.

Change it to be only defined if a cartesian-product function is not already defined.

Fixes #924.

### DIFF
--- a/stack/maxima/rtest_assessment_simpboth.mac
+++ b/stack/maxima/rtest_assessment_simpboth.mac
@@ -373,3 +373,5 @@ factorlist(-x^2-5*x+6);
 factorlist(x^3-1); 
 [x-1,x^2+x+1]$ 
 
+cartesian_product({1, 2}, {3, 4});
+{[1, 3], [1, 4], [2, 3], [2, 4]}$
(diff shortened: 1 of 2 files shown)


sangwinc:
Thanks so much for this, I really appreciate you fixing the code.  I'm going to merge this into the "dev" branch to keep master as close as possible to an official release.  This fix will then make it into the official release.  I'll close this pull request without merging it, but I've got your contribution on my dev machine.  Thanks!

PR #927, opened 2023-02-23
use nav for top menu

georgekinnear:


sangwinc:
Thanks George, that's a big improvement.

PR #935, opened 2023-03-25
Better normalise floats for cross compatibility between SBCL and GCL Maxima

cameron1729:
**Note**: This branch should resolve https://github.com/maths/moodle-qtype_stack/issues/720 and should be considered a "better" fix than https://github.com/maths/moodle-qtype_stack/pull/721 - as it allows tests to pass regardless of the specific Lisp compiler used to compile Maxima. Additionally it adds some misc CI and PHPUnit fixes.

**Background**
When using Maxima compiled under SBCL (which is used by [maximapool-docker](https://github.com/catalyst/maximapool-docker) for performance reasons), the CAS returns different values for scientific notation compared to Maxima compiled under GCL (which is what you get from Ubuntu/Debian's repositories). To see this; compare the results of the expression `scientific_notation(0.000000000001)` in Maxima compiled against GCL vs SBCL:

**GCL**
````
wget https://sourceforge.net/projects/maxima/files/Maxima-Linux/5.42.0-Linux/maxima-gcl_5.42.0-1_amd64.deb https://sourceforge.net/projects/maxima/files/Maxima-Linux/5.42.0-Linux/maxima-common_5.42.0-1_all.deb/download
sudo dpkg -i maxima-common_5.42.0-1_all.deb maxima-gcl_5.42.0-1_amd64.deb
````

Evaluate the expression:

```
 echo "scientific_notation(0.000000000001);" | maxima
```

The response is `scientific_notation(1.0E-12)`

**SBCL**
```
sudo apt-get purge maxima-common
wget https://sourceforge.net/projects/maxima/files/Maxima-Linux/5.42.0-Linux/maxima-sbcl_5.42.0-1_amd64.deb
sudo dpkg -i maxima-common_5.42.0-1_all.deb maxima-sbcl_5.42.0-1_amd64.deb  
```

Evaluate the expression:

```
 echo "scientific_notation(0.000000000001);" | maxima
```

The response is `scientific_notation(9.999999999999999e-13)` (although somehow in unit tests this becomes 10e-13).

**To see why this is a problem**
1. With SBCL Maxima  still installed, install Moodle with qtype_stack:
```
git clone git@github.com:maths/moodle-qbehaviour_adaptivemultipart.git question/behaviour/adaptivemultipart
git clone git@github.com:maths/moodle-qbehaviour_dfcbmexplicitvaildate.git question/behaviour/dfcbmexplicitvaildate
git clone git@github.com:maths/moodle-qbehaviour_dfexplicitvaildate.git question/behaviour/dfexplicitvaildate
git clone git@github.com:maths/moodle-qtype_stack.git question/type/stack
```
2. Update `config.php` with the following lines:
```define("QTYPE_STACK_TEST_CONFIG_MAXIMAVERSION",       "5.42.0");
define("QTYPE_STACK_TEST_CONFIG_MAXIMACOMMAND",       "maxima");
define("QTYPE_STACK_TEST_CONFIG_CASTIMEOUT",          "10");
define("QTYPE_STACK_TEST_CONFIG_MAXIMALIBRARIES",     "stats, distrib, descriptive, simplex");
define("QTYPE_STACK_TEST_CONFIG_CASPREPARSE", "true");
define("QTYPE_STACK_TEST_CONFIG_CASRESULTSCACHE",     "db");
define("QTYPE_STACK_TEST_CONFIG_PLOTCOMMAND",         "");
define("QTYPE_STACK_TEST_CONFIG_CASDEBUGGING",        "0");
define("QTYPE_STACK_TEST_CONFIG_PLATFORM",            "linux");
define("QTYPE_STACK_TEST_CONFIG_MAXIMACOMMANDSERVER", "");
define("QTYPE_STACK_TEST_CONFIG_MAXIMACOMMANDOPT",    "");
```
3. Setup PHPUnit and run the following:
```
vendor/bin/phpunit --testsuite=qtype_stack_testsuite --stop-on-failure --filter='test_scientific_notation'
```

The same error will be observed when using maximapool-docker, as that uses SBCL Maxima.

There is some pre-existing code in `test_base::prepare_actual_maths_floats` to handle situations where different Maxima setups return different values. So the most logical thing to me is to extend that code to normalise the coefficient of numbers using scientific notation. Which is what this patch does. This way the test pass when using GCL (which is what GHA is using) or SBCL Maxima.

**Testing**
1. Checkout this branch for `qtype_stack`:
```
cd question/type/stack
git remote add catalyst git@github.com:catalyst/moodle-qtype_stack.git
git fetch catalyst phpunit-fixes
git checkout phpunit-fixes
```

2. Run `vendor/bin/phpunit --testsuite=qtype_stack_testsuite --stop-on-failure --filter='test_scientific_notation'` on both SBCL and GCL Maxima and ensure there are no failures. No change to `config.php` is needed. Just change Maxima like so:

For GCL:
```
sudo apt-get purge maxima-common
sudo dpkg -i maxima-common_5.42.0-1_all.deb maxima-gcl_5.42.0-1_amd64.deb
```

For SBCL:
```
sudo apt-get purge maxima-common
sudo dpkg -i maxima-common_5.42.0-1_all.deb maxima-sbcl_5.42.0-1_amd64.deb
```

You can check which version of Maxima your system is using like so: `echo "build_info();" | maxima`

### DIFF
--- a/doc/docslib.php
+++ b/doc/docslib.php
@@ -142,7 +142,7 @@ function stack_docs_render_markdown($page, $preprocess = true) {
         // Don't process the auto-generated answer test output.
         $page = stack_maths::pre_process_docs_page($page);
     }
-    $page = format_text($page, FORMAT_MARKDOWN, array('filter' => false, 'noclean' => true));
+    $page = format_text($page, FORMAT_MARKDOWN, array('filter' => false));
     $page = stack_maths::post_process_docs_page($page);
     return $page;
 }
--- a/.github/workflows/moodle-ci.yml
+++ b/.github/workflows/moodle-ci.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   test:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-22.04
 
     services:
       postgres:
@@ -30,31 +30,43 @@ jobs:
     strategy:
       fail-fast: false
       matrix: # I don't know why, but mariadb is much slower, so mostly use pgsql.
+        # We use a mix of SBCL and GCL, but mostly prefer SBCL as it is faster.
         include:
           - php: '8.0'
             moodle-branch: 'master'
             database: 'pgsql'
+            maxima: 'SBCL'
           - php: '7.4'
             moodle-branch: 'MOODLE_401_STABLE'
             database: 'pgsql'
+            maxima: 'SBCL'
           - php: '7.4'
             moodle-branch: 'MOODLE_400_STABLE'
             database: 'pgsql'
+            maxima: 'GCL'
           - php: '7.3'
             moodle-branch: 'MOODLE_311_STABLE'
             database: 'pgsql'
+            maxima: 'SBCL'
           - php: '7.3'
             moodle-branch: 'MOODLE_310_STABLE'
             database: 'pgsql'
+            maxima: 'SBCL'
           - php: '7.2'
             moodle-branch: 'MOODLE_39_STABLE'
             database: 'mariadb'
+            maxima: 'SBCL'
 
     steps:
-      - name: Install required libraries
+      - name: Install Maxima (${{ matrix.maxima }})
         run: |
-          sudo apt-get install gnuplot maxima maxima-share texinfo
-          maxima --list-avail
+          maxima="${{ (matrix.maxima == 'SBCL' && 'sbcl') || 'gcl' }}"
+          wget http://mirrors.kernel.org/ubuntu/pool/main/r/readline/libreadline7_7.0-3_amd64.deb \
+               https://sourceforge.net/projects/maxima/files/Maxima-Linux/5.42.2-Linux/maxima-common_5.42.2-1_all.deb \
+               https://sourceforge.net/projects/maxima/files/Maxima-Linux/5.42.2-Linux/maxima-${maxima}_5.42.2-1_amd64.deb
+          sudo apt-get install libtinfo5
+          sudo dpkg -i libreadline7_7.0-3_amd64.deb maxima-common_5.42.2-1_all.deb maxima-${maxima}_5.42.2-1_amd64.deb
+
           echo "diff(x^2,x);" | maxima
           echo "build_info();" | maxima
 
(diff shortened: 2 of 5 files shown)


cameron1729:
I've added another commit which updates the GHA workflow to test against SBCL Maxima as well as GCL Maxima; hopefully this can help identify any further incompatibilities in the future.

timhunt:
Please don't add so many different workflow runs. It is not necessary to run a full matrix of possibilities, and given that GitHub is gifting us free CPU time for our CI, I think it is rude.

It is enoguh to just sample the matrix, so try randomly alocating SBCL ro GCL to the different runs - making sure that it isan't alwasy a particular Maxima with a particular Lisp, and vice versa.

Oh! we seem to have lost any mysql runs. When did that happen?

cameron1729:
Thanks @timhunt 

> Please don't add so many different workflow runs. It is not necessary to run a full matrix of possibilities, and given that GitHub is gifting us free CPU time for our CI, I think it is rude.

What do you think about testing GCL and SBCL for master only? That way we only add one extra workflow run.

timhunt:
Given we haev a few different runs, for diffeerent Moodle versions, I woudl just make one or two of those runs use the other LISP. (If one one is faster than the other, use more of that. I think that is why we have more Postgres than MySQL.)

aharjula:
Frankly, there is absolutely no need to test the same DB+Moodle combo with multiple LISP+Maxima-version combos. Any breakage on either side should be easily visible, no matter what the other side has. The code that connects these is such that the other side will not be fixing whatever went wrong on the other end, and therefore no pairing should be special.

Given the style of the DB-queries we have, testing different DBs on the same Moodle is also somewhat questionable. As basic database abstractions are surely already well-tested in Moodle and we really are not pushing any boundaries there.

However, testing different LISPs for different Maxima-versions is something worth doing if possible. Maxima is a much faster-moving target, and we rely on it for pretty much everything. Now if we could get some newer Maximas...

doc/docslib.php
sangwinc, line 1: Ok, thanks for cleaning this up!

sangwinc:
Could I ask you to consolidate some of the combinations a little as suggested?

Also, could you send a pull request to the dev branch, not to master?  We'll then be able to test this ourselves, and make any further changes before we include this in the next release.

Chris

cameron1729:
Thanks @sangwinc 

I've updated the workflow steps to mostly use SBCL. I found it's faster for all branches. Which can be seen [here](https://github.com/catalyst/moodle-qtype_stack/actions/runs/4540008523) in this run in our fork. I picked MOODLE_400_STABLE to use with GCL as it completes reasonably fast. It seems that sometimes SBCL can finish a lot faster (e.g., the MOODLE_39_STABLE branch took 1h 41m with GCL vs 1h 8m with SBCL). Although in [another run](https://github.com/catalyst/moodle-qtype_stack/actions/runs/4539159015) the timings were similar (SBCL was still faster though).

I also updated the workflow to use Ubuntu 22.04 as 18.04 is [going away soon](https://github.com/actions/runner-images/issues/6002). PR has also been changed to merge into `dev`.

Let me know if there's anything else. Looking forward to having this merged :)

cameron1729:
@sangwinc - did this ever make its way in to a release? As best I can tell this is still not present in master.

sangwinc:
@cameron1729 I do appreciate you contributing here - thanks.

Yes, we did update Ubuntu
https://github.com/maths/moodle-qtype_stack/blob/master/.github/workflows/moodle-ci.yml#L7

Your addition of different versions of Maxima was much appreciated and is in master here
https://github.com/maths/moodle-qtype_stack/blob/master/.github/workflows/moodle-ci.yml#L63

I know catalyst are important commercial partners.  If you have a particular setup you would like us to include in the CI then please add it.  Edinburgh has done so here
https://github.com/maths/moodle-qtype_stack/blob/master/.github/workflows/moodle-ci.yml#L45

I may have missed something you suggested, and if so I'm sorry! Please send me another pull request.  This kind of CI testing is an ongoing battle...

Chris

PR #947, opened 2023-04-27
Update symbols-stack.json

Bjoern-Ge:
I added 3 codes that I regularly meet with my students as they are part of the German keyboard on the iPad


sangwinc:
Thanks very much for these helpful suggestions!  The place to fix them is in our shared code on NUMBAS, which I've done.

PR #948, opened 2023-04-28
Fix PHP deprecated error in inputbase.class.php

AnupamaSarjoshi:
Hi Chris, 

We found some PHP deprecated error when updated to PHP 8.1. Hence have done a small change to fix the same in stack/input/inputbase.class.php. Could you please review?

I have done the change assuming that the minimum supported PHP version for STACK is PHP 7.4 or later. I hope that is OK.

Thanks,
Anupama.


sangwinc:
Thank you that's much appreciated!  It prompts me to add in PGP 8.1 to the CI, which I'll do once this is merged.

Issue #951, opened 2023-05-16
stack_jxg.bind_list_of


Famondir:
Hello all,
we are running STACK-Version 2023010400 and in our documentation there is mentioned the function `stack_jxg.bind_list_of`. Unfortunately I get an error (in the console in developer view) that says there is no such a function. The fuction `stack_jxg.bind_point` works just fine.

Do you have any clue what could be the reason or can you advice me what further information I can provide you / things I could try?

Best regards
Simon

sangwinc:
Very sorry about this.  It will be fixed in the next (imminent) release.

PR #961, opened 2023-05-24
Add extra example to JSXGraph.md

LukeLongworth:
Added a detailed example on region shading as a way to explore custom binding functions.

I acknowledge that this may work better as a case study or as its own page in the docs - it is rather long - but I figured I would submit it here first. If you think it SHOULD be moved, I would add some suggestions for how to introduce randomisation to this question and discuss the difficulties with this type of question when the regions aren't uniform like this.

### DIFF
--- a/doc/en/Authoring/JSXGraph.md
+++ b/doc/en/Authoring/JSXGraph.md
@@ -164,6 +164,7 @@ To parse and manipulate it you can use STACK's custom JSON parsing functions:
     tmp:stackmap_set(tmp,"z",x*y);   /* ["stack_map", ["x", 4], ["y", 3], ["z", 12]] */
     json:stackjson_stringify(tmp);   /* "{\"x\":4,\"y\":3,\"z\":12}" */
 
+Another more complicated example of custom binding is shown at the bottom of the page.
 
 ## Convenience tools for building graphs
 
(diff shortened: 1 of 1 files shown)


aharjula:
Looks nice. Just two things:
 
1. The upcoming 4.4.3 release will bring a new JS system where the code executes in a separate context and inputs get synchronised only on "change", therefore when you write directly to `ìnput.value` you will also need to trigger a change event to make the system notice things. Basically, do [this](https://github.com/maths/moodle-qtype_stack/blob/c89702e4039eaed2f80274d8b423e0fcf13420f3/doc/en/Authoring/JSXGraph.md?plain=1#L139-L141). Triggering those events in the pre 4.4.3 system should not cause damage.
2.  I would really like people to start using `/* block comments */` instead of `// comments` in examples, as those are safe from various import/export issues. Of course, I know that all the examples are not using those...

Otherwise without testing that looks sane, and is one of the cases where actual serialiser/deserialiser-based implementation using the new [`custom_bind()`](https://github.com/maths/moodle-qtype_stack/blob/c89702e4039eaed2f80274d8b423e0fcf13420f3/corsscripts/stackjsxgraph.js#L222) might not make sense, however, the last one with the matrix would work better with custom bind. Note that 4.4.3 will [finally](https://github.com/maths/moodle-qtype_stack/issues/951) bring the new javascript binding functions mentioned in the docs but missing from the system, and `custom_bind()` is now the general way of doing binding without having to deal with events if one only wants to tune the representation at the (de)serialisation level.

LukeLongworth:
> Looks nice. Just two things:
> 
> 1. The upcoming 4.4.3 release will bring a new JS system where the code executes in a separate context and inputs get synchronised only on "change", therefore when you write directly to `ìnput.value` you will also need to trigger a change event to make the system notice things. Basically, do [this](https://github.com/maths/moodle-qtype_stack/blob/c89702e4039eaed2f80274d8b423e0fcf13420f3/doc/en/Authoring/JSXGraph.md?plain=1#L139-L141). Triggering those events in the pre 4.4.3 system should not cause damage.
> 2. I would really like people to start using `/* block comments */` instead of `// comments` in examples, as those are safe from various import/export issues. Of course, I know that all the examples are not using those...
> 
> Otherwise without testing that looks sane, and is one of the cases where actual serialiser/deserialiser-based implementation using the new [`custom_bind()`](https://github.com/maths/moodle-qtype_stack/blob/c89702e4039eaed2f80274d8b423e0fcf13420f3/corsscripts/stackjsxgraph.js#L222) might not make sense, however, the last one with the matrix would work better with custom bind. Note that 4.4.3 will [finally](https://github.com/maths/moodle-qtype_stack/issues/951) bring the new javascript binding functions mentioned in the docs but missing from the system, and `custom_bind()` is now the general way of doing binding without having to deal with events if one only wants to tune the representation at the (de)serialisation level.

Thanks for the comments!

1. Good to know about the upcoming changes. We use this type of thing in a few different places, so I'll have to make those changes when we update. I hadn't really thought about it, but I had been referencing our local version of the docs when writing this, not the latest version, so I missed that entirely. I'll have a play with that and then make the edits later. 

2. Ah! This is how I usually comment, but I copied the above example this time... I'll keep that in mind too. 

I'm excited for the new custom_bind feature, I thought I had read something about that at some point, but wasn't sure where. Once we get this on our system (won't be until end of the year at earliest) I would happily experiment with this and write up a similar example. 

I'm very new to using GitHub. What is the most sensible way for me to make these edits? Should I close this Pull request, make changes, commit, and open a new Pull request?

aharjula:
Not a GitHub expert on pull requests as I tend to just push to the dev, but below your comment, it says, "Add more commits by pushing to the patch-5 branch on LukeLongworth/moodle-qtype_stack." I do believe that works, if not then the close and redo thing probably does the trick.

In any case, thank you for this example.

sangwinc:
Luke, we're about to release the next version so I suggest we release, and then we update these new examples for version 4.4.3.  Is that ok?   So, I won't merge this now.

PR #964, opened 2023-05-26
Increase Node and Standard feedback box sizes to two for usability

AnupamaSarjoshi:
Hi Chris,

When we create/edit a Stack question, the Node and Standard feedback boxes seem to very small, of strictly one line. Changing the number of rows to two should increase the usability. (Also, we recently received a complaint by staff member regarding the feedback boxes being very small for some of the question types.)

Could you please review the patch? We assume this would be an useful change.

Thanks,
Anupama


sangwinc:
Thanks, that looks like an improvement.
Chris

PR #970, opened 2023-06-12
Created a Vector_Calculus section in the docs

LukeLongworth:
This is still a bit rough, but I think it's in a form that is workable and ready for other STACK developers to have a look at.


sangwinc:
Many thanks Luke!

Issue #974, opened 2023-06-13
Hide test options 


m-r-k:
I propose we hide the test options input field for answer tests who do not need them

sangwinc:
That's great thanks!  Much appreciated simple improvement.

PR #976, opened 2023-06-14
Fixes crash related to #974 (null for PRT answer option)

geoo89:


sangwinc:
Many thanks for fixing this.

PR #978, opened 2023-06-14
Update vectorcalculus.mac

LukeLongworth:
Added author details, fixed typo, added error detection to curl function vars field, added one more question test (u as row vector to div).


sangwinc:
Thanks Luke,
Chris

PR #988, opened 2023-07-10
Comments on handling decimal separation

PM84:


sangwinc:
Thanks for the comments Peter.

PR #991, opened 2023-07-13
Fix PHP8.1 deprectation error for trim in question.php

AnupamaSarjoshi:
Hi Chris,

This a one more small fix to resolve the PHP8.1 deprecation error. Could you please review?

Deprecated: trim(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/html/moodle/question/type/stack/question.php on line 1495

Thanks,
Anupama


sangwinc:
Thank you!

Issue #993, opened 2023-07-18
Proposal: change STACK options to cascading options


sangwinc:
We have a lot of options which don't get set very often, and many of these are what I think of as "aesthetic" rather than as "content".  For example, lots of the display preference options such as the logic symbol (word vs symbol).
https://github.com/maths/moodle-qtype_stack/blob/master/stack/options.class.php#L64

Options such as the shape of brackets
https://github.com/maths/moodle-qtype_stack/blob/master/stack/options.class.php#L112
normally is an aesthetic preference, but of course may need to be set for some applications.

The proposal is to add a new value for every option of "default".

1. New questions start with the "default" value for the option.
2. The "default" value of the option queries the site default when the question is compiled, and the option used.
    This might be a simple single line here:  https://github.com/maths/moodle-qtype_stack/blob/master/stack/options.class.php#L176
3. We plan a text-based (YAML?) question authoring format.  I expect users will only need to set options which differ from the default in any text-based authoring system.  I don't see a problem with a text-based authoring system without fixing this issue, but this issue confirms the design.

Note to self: changing the site default will need to result in clearing the question compilation cache.  Of course we almost never actually change site default options and so this performance hit will not really be a significant issue.

This is particularly useful for options such as the decimal separator proposed in issue #789.  When we share questions we almost certainly want to import/export materials using the "default" option.  Continental users will not want my questions with a British "." as the decimal separator and the certainly don't want to edit them!

(We had this feature in STACK v1.0 at the start of the century but it got lost, sorry!)

aharjula:
Now if we only had this at the course/category level in Moodle instead of the site level. So that all the electrical engineers could switch to using the `j` in all questions and everyone else could simply keep `i` in play. Moodle does not provide a system for that, but I wonder if we could have one of those Moodle "blocks" that one could plug into a course or category, override configuration there and then we could look for those in the hierarchy of the site and pick the values from the nearest one, or from the whole chain.

As to recompiling, that depends on the options themselves. Most of them are just display options and are present in the preamble that gets generated during use, just like setting active the language and seed and thus have no role in the code coming out of the compilation. So simply checking which options get changed will probably save us from doing that. Overall, all options should be clearly defined and classified in some meta-data that can be processed in code so that we know which ones affect specific parts of logic. And I think that we do not currently have any options that would affect the compilation result.

sangwinc:
Thanks for the clarification on caching Matti.

[...]

PR #1008, opened 2023-08-08
Remove styling from jsxgraph.min.css

georgekinnear:
Rather than setting background and border here, I think it would be better done by putting a class on the iframe object, so that it can be customised by the Moodle theme.


timhunt:
Surely jsxgraph.min.css is an auto-generated file, so we should not be editing it? (Sorry, I don't acutally know much about jsxgraph, just general prinicples - which might not be relevant here.)

georgekinnear:
Yes, fair enough, but I couldn't see the non-min'd file anywhere!

aharjula:
The CSS comes directly from the JSXGraph release: https://github.com/jsxgraph/jsxgraph

It applies only inside the iframe and I doubt you can actually apply any Moodle side themes to the interior of the iframe.

Should you really want to override the CSS used inside the iframe you can use the special attribute to load any CSS you want `[[jsxgraph overridecss="url"]]`, `overridejs` is equivalent and allows you to override the JSXGraph library itself. With that, you can actually replicate the old behaviour where we did not load any CSS for `[[jsxgraph]]` at all.

In any case, I don't believe this pull request would be beneficial. At least it does not match my internal view of the STACK-JS  system and truly separate iframes that are truly detached from the surrounding VLE including styling...

timhunt:
@georgekinnear do you have any connection into the JSXgraph community? Might be worth finding a tactful way to ask them "why is your stying so opinionated? Could it be more neutral be default?"

LukeLongworth:
At UC we run a script at the top of all of our JSXGraph questions to set default behaviour and styles. It's not the best solution, and I would welcome any native ability to set site- or course-wide default styles.

aharjula:
Ok. This will be a long one...

First, sorry for messing things up for UC, the iframe-based separation of logic that secures our JavaScript and forms the backbone of STACK-JS probably breaks all theming you have built. After some opinions, the possible routes forward are outlined below. As to setting defaults, you should probably bring things up in #993. I have a pretty good idea of how the stylesheets could be dealt with if we had a way of having such settings. Having suitably scoped settings is the problem, though.

Secondly, my opinionated opinion is that STACK has no right to override the default style of something like JSXGraph. Who are we to do so? We should really focus on ensuring that its default is respected as the default so that all documentation and examples follow a similar style as the source system. This opinion comes from a person that for quite some time included JSXGraph into STACK without any styles whatsoever but has since seen the bad form in my own way. However, we may naturally enable the authors to override whatever they want and should absolutely provide means to do so, as who are we to say what the authors should author?

Now then a long list of ways to control the styles:
 1. The already mentioned `[[jsxgraph overridecss="..."]]` is, of course, one, maybe inconvenient but probably the most obvious in its use. Should one manage to get ones stylesheet file into the [/corsscripts/](https://github.com/maths/moodle-qtype_stack/tree/master/corsscripts) directory, one could simply write `[[jsxgraph overridecss="cors://mystyle.css"]]` and that would map to that directory on the local install. Note that including any external CSS-files into the iframe will require that those files are being served with a particular CORS header. Unsurprisingly we have a particular logic for serving stuff using a PHP script inside that /corsscript/-directory.
 2. Second option is to use the named-version system that was built just in case JSXGraph ever would have a massive API change, and we would need to declare what version we use, but which could be used, for example, to freeze JSXGraph version for something like HELM, it declares the script and style content [here](https://github.com/maths/moodle-qtype_stack/blob/ec5cc8e6dffb2b915c2fa682870e2c1ef8af396a/stack/cas/castext2/blocks/jsxgraph.block.php#L31-L41), and you would use it like this `[[jsxgraph version="local"]]`. Should something like HELM distribute its styles (etc.) with STACK this might be the way to go. Obviously, no one is using this yet, but it is there and allows declaring multiple site-level customisation versions.
 3. Then we have the `[[style]]` block that is currently still something with no real documentation but which acts as part of the whole `[[iframe]]` block family of which `[[jsxgraph]]` is part of. It allows one to write CSS inside the block so that it gets applied to that block. For example, writing this is entirely acceptable:
 ```
 [[jsxgraph]]
 ....
 [[style]]
.jxgbox  {
border-style:solid;
border-width:4px;
border-color:#356aa0;
}
[[/style]]
....
[[/jsxgraph]]
 ```
 And so is this:
  ```
 [[jsxgraph]]
 ....
 [[style href="..."/]]
....
[[/jsxgraph]]
 ```
 You can use basically, all the attributes one typically uses with the style element or link element in normal HTML authoring with that `[[style]]`-block, i.e. providing separate `media` versions of style is possible, even though probably putting those media selectors inside the stylesheet is probably simpler.
 4. The `[[style]]` block can naturally be mixed with the `[[include]]`-block, but then you will need to deal with all the compilation-related issues, i.e. unless using that "href"-version, those style settings will be frozen and no matter how you change the include target those values are not flowing into production unless one recompiles the question. But yes you could have:
 ```
 [[jsxgraph]]
 [[include src=".../our-common-jsxgraph-logic-and-styles.txt"/]]
 [[/jsxgraph]]
 ```

Obviously, some of those are very much tools for single graph level customisation, and others for larger scale operations and it is up to the use case to decide which ones to apply and whether one wants to stack multiple ways on top of each other.

aharjula (commented):
Does not break anything other than consistent style, but I am not in favour of this.

georgekinnear:
Thanks for this Matti.

I agree with you on principle about not overriding the JSXGraph defaults. It's only an issue for me because the appearance has changed quite noticeably since the upgrade - the content of my questions has not changed, but now the plots have white backgrounds and rounded borders. So it seems from my perspective like a change in behaviour from the upgrade.

(Our demo site has yet to be updated so the questions there still appear without backgrounds and borders, e.g. here: https://stack-demo.maths.ed.ac.uk/demo/course/view.php?id=18  - it seems like there is no definition of the .jxgbox style being loaded.)

It's helpful to know all those different approaches you suggest. I think the 3rd one is the simplest in my case, and I can achieve the effect of putting my questions "back to normal" by inserting this code at the start of any `[[jsxgraph]]`:

```
[[style]]
.jxgbox  {
  border-style:none;
  background: none;
}
[[/style]]
```

As Luke says, some ability to define/apply styles would be useful - we've discussed some ideas about that over email and perhaps the next step is for me to open an issue here to gather more thoughts.

PR #1013, opened 2023-08-15
STACK: Instant validation not presented to screen-reader users #661466

sangwinc:


timhunt:
Ah! I was intending to comment on this. I have not tested this myself, but are we sure it is right to always add the 'aria-live' => 'assertive' attribute.

What happens in the (less common) case where you have several questions on one page of a quiz, and you click the 'Check' button on one question (or review the attempt as a teacher).

Does it assertively read out all the instant validation from all the inputs? That would be bad.

Or, does this only take effect when the ajax code updates the validation when one input has changed?

@KhoaNguyen, did you test that?

PR #1032, opened 2023-09-29
Update Authoring_GeoGebra.md

guidopinkernell:
removed info about defining editor mode to html globally, instead added info about defining editor mode to html locally where it applies.


sangwinc:
If your're sure Guido.  From bitter experience I (and @KZerva ) have had to debug a lot of questions caused by authors editing things with the default editor....

PR #1078, opened 2023-12-01
Updated JSXGraph.md

LukeLongworth:
Changed comment format from // to /**/ and added two examples to the bottom of the page on custom binding. Also removed the suggestions to do `stateInput.style.display = 'none';` because a) the (in my opinion) much better* method of `<p hidden="">[[input]]</p>` is already given earlier in the page, and b) I'm not convinced this method still even works with STACK-JS. 

*I think it's better because if you use `style.display = 'none'` then the input box is visible on page load and then disappears once everything else has loaded.


aharjula:
Yes STACK-JS did indeed make it impossible to simply modify the visibility of an input like that. And elimination of `//` comments in any JavaScript inside Moodle will always be welcome.

The sector shading example is interesting and would probably map to normal `custom_bind` with a "handle object", like the hidden padding point in this [upcoming documentation](https://github.com/maths/moodle-qtype_stack/blob/dev/doc/en/Authoring/Advanced_JSXGraph.md) (requires 4.4.7 for the `register_object`-calls to work in that example). The `custom_bind` does not need to relate to the actual objects it has been given to track, those are just the objects that when updating trigger the binding logic.

Handle objects are also handy if you attach additional event handlers to your graph, and it becomes unclear in which order those might get executed. In an [example on the Zulip side](https://stack-assessment.zulipchat.com/#narrow/stream/384527-visualization/topic/Bind.20properties.20of.20Point.20from.20JSXGraph.20diagram.20in.20Stack) we use the actual bound objects like handle objects to ensure that the event hander logic happens after things have been done.

In any case, I support merging this in this form. Probably will be rearranged more when we start having more examples.

sangwinc:
Thanks Luke, a much appreciated contribution.

PR #1084, opened 2023-12-06
Update validators.mac

LukeLongworth:
Added validate_all_one_letter_variables which invalidates any answer with variable names > 1 character (including Greek letters - not intended, but also likely to not be a problem).


sangwinc:
Thanks Luke, that's much appreciated.

PR #1112, opened 2024-02-14
four typos

Mat-Ge:


sangwinc:
Thanks, that is much appreciated.

PR #1113, opened 2024-02-14
Fix PHP8.2 deprecated error for trim in questiontestreport.php

AnupamaSarjoshi:
Hi Chris,

We have noticed some PHP deprecation errors related to PHP8.2 trim function on 'Analyze responses' page (questiontestreport.php). Could you please review the fix?

Deprecated: trim(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/html/moodle/question/type/stack/questiontestreport.php on line 141

Thanks,
Anupama


sangwinc:
Much appreciated, thank you!

PR #1130, opened 2024-03-08
Update Differential_equations.md

shinuito:
Add an example to the docs of manipulating Laplace Transforms with Maxima.


sangwinc:
Thanks - that's a really helpful addition.

PR #1134, opened 2024-03-13
Update validators.mac

LukeLongworth:
Fixed a typo in the `validate_interval_syntax` feedback; it previously used square brackets for an open interval


sangwinc:
Thanks Luke.

PR #1154, opened 2024-03-26
Iss973 - Format setting for question note

EJMFarrow:
**Contains a non-reversible DB change**

This closely follow changes from the addition of previous fields.
Main point of note is question.php->get_question_summary() which bypasses URL re-writing for images/files as we don't have the required question attempt info. This is the function usually used for retrieving question notes so anyone using images will likely be disappointed but that's not necessarily a bad thing...


sangwinc:
Thanks @EJMFarrow this looks great - time we put this into dev and moved on.

Issue #1058, opened 2023-11-13
Pedantic display: Variable matrix and compact validation


timlowe:
Hi,

When using "variable matrix" input and compact validation, the validation does not  nicely align with the input field. (It does align for fixed-size matrix.) See attached.

Also, can the matrix input be made to appear inline with the preceding text (as it does for numerical entry)? 
The question code has  \( \mathbf{A} = \) [[input: ans1]], but the preview shows a line break.

(I realise this is pedantic - just trying to make the question look appealing!)

Tim

![variable matrix](https://github.com/maths/moodle-qtype_stack/assets/5654528/ca399336-7477-4f89-8eab-c1fb3877d05a)

LukeLongworth:
Ah, we have had this problem too. My modern fix is to put the A= and the input into adjacent cells of a table without borders, but this isn't particularly elegant. 

If I'm remembering correctly, this is (partially) mitigated if you don't put this in <p> tags. When the question page is generated I think it inserts some of its own <p> tags which causes the line break. My earlier fix for this problem was putting it all in a <span> tag. Note: I am not a developer and this was a while ago!

georgekinnear:
I think I've seen this issue before too - the solution is to edit your HTML so that the "A = ..." is in a `<div>` rather than a `<p>`.

(The issue is with HTML, where "display" objects cannot appear inside "inline" objects - the matrix input is a "display" object so it cannot sit inside a `<p>` which is an "inline" object: https://stackoverflow.com/questions/8397852/why-cant-the-p-tag-contain-a-div-tag-inside-it)

timlowe:
Thanks George! Yes, using <div> overcomes the line-break issue. but the misaligned compact verification feedback remains!

(Luke  - I'd rather not use a table, as that makes life difficult for screen-reader users!)

georgekinnear:
I'm not sure if there's a better way other than to put the `[[validation:ans1]]` tag inside a `<div>` that you tweak to get it in the right position - e.g.

```
<div>\(A=\) [[input:ans1]]
    <div style="display: inline-block; vertical-align: top;">[[validation:ans1]]</div>
</div>
```

That's still not perfect, so you maybe want to add `margin-top: 0.2 em` to the style (but you could spend forever tweaking that number to make it line up perfectly, I fear!)

sangwinc:
George, are you suggesting we add this <div> style round the validation message as a default?  If so, what happens to people who've also added it manually as you suggest above?

[...]

PR #1155, opened 2024-03-26
add vertical alignment to inline block feedback

georgekinnear:
Proposed fix to issue #1058


sangwinc:
Thanks @georgekinnear.

PR #1163, opened 2024-04-09
Behat additions

EJMFarrow:
Add basic Behat tests for each input type and a simple JSX Graph question.
Add a test of back up as part of basic plugin coverage.
Extra test added to xml_import.feature as part of Iss973.


sangwinc:
Thanks @EJMFarrow.

PR #1164, opened 2024-04-09
WIP - Iss1053 - Instant feedback doesn't follow forced language

EJMFarrow:
Hi Matti, I could do with you looking this over, please. I've included the language in a hidden input in the rendered question and now pass it back on the validation request. The language is also saved in the step data when the answer is submitted.

Will discuss test coverage with Chris.


stack/input/inputbase.class.php
```diff
@@ -1485,7 +1492,7 @@ public function maxima_to_response_array($in) {
     public function replace_validation_tags($state, $fieldname, $questiontext) {
 
         $name = $this->name;
-        $feedback = $this->render_validation($state, $fieldname);
+        $feedback = $this->render_validation($state, $fieldname, null);
```
aharjula, line 1495: Maybe a comment here to tell why we use `null` as `$lang`; it makes sense to me that as this path is always executed with a sensible language, we do not need to override it and do that switching around, but why might not be so clear to everyone. Also are we absolutely sure that PHP syntax will never change so that the ifs on line 1322 and 1373 above will not start thinking null as true, maybe those ifs should check $lang !== null`... Same for all the other checks in all the other input types.

renderer.php
```diff
@@ -141,6 +141,14 @@ public function formulation_and_controls(question_attempt $qa, question_display_
...
         }
 
+        // ISS1053 - Output current language
+        // to be returned by validaion request and also stored in
+        // question step data on submission.
+        $fieldname = $qa->get_qt_field_name('step_lang');
+        $currentlang = current_language();
+        $result .= html_writer::empty_tag('input', array('type' => 'hidden',
+                    'name' => $fieldname, 'value' => $currentlang));
+
```
aharjula, line 151: Yes, this ties to the UI language, not to the question language, which might be different, i.e., it's good that you did not mix `stack_multilang` into this.. I wonder, though, if this is always the same as the `<html>`-tags lang`-attribute. STACK-JS already syncs that over to the iframe, and if this would ever be different, then it would need to sync this instead.

amd/src/input.js
```diff
@@ -430,11 +431,16 @@ define([
      */
     function initInputs(questionDivId, prefix, qaid, inputs) {
         var questionDiv = document.getElementById(questionDivId);
+        var language = null;
+        var langInput = document.getElementsByName(prefix + '_step_lang');
+        if (langInput.length > 0 && langInput[0].value) {
+            language = langInput[0].value;
```
aharjula, line 437: Does this work? Basically, why do I see `null` in my requests when doing preview?

Basically, started wondering after I noticed that things work in preview-popup just fine but if I change the language of Moodle in the background in another tab then the preview validation also changes language, which it should not do if this parameter is the defining factor and comes from the document inside that preview which does not change.

aharjula: Naturally, I assume that the minified scripts matched this.

renderer.php
EJMFarrow: Iframe block is doing
`$code .= '<html xmlns="http://www.w3.org/TR/xhtml1/strict" lang="' . 
            stack_get_system_language() . '">';`
which is essentially the same thing. Is that what you're referring to or is there something else I'm missing?

amd/src/input.js
EJMFarrow: My bad. Did a last minute update to the element name to make it more unique and messed it up. Hopefully fixed now.

renderer.php
aharjula: Nothing, I am hallucinating something; I had a memory that I actually read that value from the VLE DOM. But then that would have been slow and pointless.

amd/src/input.js
aharjula: Now it works as expected.

PR #1181, opened 2024-05-03
Behat fix

EJMFarrow:
Issue was newer versions of Moodle default to async backup - this breaks the Behat tests. I had to set the config (and then oddly go into the admin page and save it to make it stick but oh well).

Language issue on the unit tests was user setting taking priority over config setting. I now blank the user setting to reinstate previous behaviour.


sangwinc:
Thanks so much @EJMFarrow for sorting this out.  Much appreciated.

PR #1190, opened 2024-05-17
Parson's v2: Extension to grids and groupings

smmercuri:
- The `parsons` block is now extended to handle multiple columns (for grouping questions) and grid layouts (for matching and other grid-style questions) through the use of `columns` and `rows` block header parameters. The original proof Parsons version of v4.5.0 is obtained by specifying neither `columns` nor `rows`.
- Update the `Sortable.css` styles for grid styling, which is separate to the proof Parsons styling.
- Add `matchlib.mac` Maxima library containing helper and required functions for grouping and grid questions using the `parsons` block. These help to translate between Maxima arrays and relevant JSON structure, as well as some useful assessment functions.
- CSS bug fix for proof assessment display
- Documentation for Matching problems with examples.


sangwinc:
Thanks @smmercuri, looks great.

PR #1191, opened 2024-05-21
Add trim to fix filename path returned by zip::getNameIndex function

AnupamaSarjoshi:
Hi Chris,

In some of the PHP versions, we see the following warning for stack questions with images. The function zip::getNameIndex() is returning the filename with a leading '/', thus setting the incorrect filepath, resulting in images not being displayed.

Error: 
Warning: file_put_contents(/efs/moodledata/stack/plots//stackplot-9809-1-3924935062-16894435.svg): Failed to open stream: No such file or directory in /var/www/html/moodle/question/type/stack/stack/cas/connector.server.class.php on line 86

Could you please review the fix?

Thanks,
Anupama


sangwinc:
Many thanks,
Chris

PR #1197, opened 2024-06-03
Automatic resizing of grid-items in `parsons` block for matching problems.

smmercuri:
Add auto-resizing functionality of elements with `grid-item` and `grid-item-rigid` classes, which appear when using the `parsons` block with non-null `columns` parameter.

This resizes the height of all headers, indices and draggable items to the maximum height of inner content across all such items. When horizontal orientation is used, this additionally resizes the widths of all such items.

The parameters `item-height` and `item-width` are still available to authors as override options, which will essentially disable automatic resizing and use the supplied values as height and/or width.


sangwinc:
@smmercuri, these changes were merged into the v4.6.0 branch, so I'll close this pull request without merging them to dev a 2nd time.  Thanks for fixing this.

Issue #940, opened 2023-04-06
Only the terminal node of a PRT shows its feedback


alcarola:
I created a multilanguage multipart question with a multistep PRT (attached). 

My problem is that only the terminal node of the PRT displays its feedback. I want all the nodes to display their feedback. See the attached testing protocol (and the question preview behaves the same, only displaying the feedback of the terminal node).

[Testing question Nonlinear algebraic system [FI_EN_SV].pdf](https://github.com/maths/moodle-qtype_stack/files/11171649/Testing.question.Nonlinear.algebraic.system.FI_EN_SV.pdf)

[questions-NumMat 2023-Nonlinear algebraic system [FIENSV]-20230406-1735.xml.zip](https://github.com/maths/moodle-qtype_stack/files/11171661/questions-NumMat.2023-Nonlinear.algebraic.system.FIENSV.-20230406-1735.xml.zip)

aharjula:
I think we have seen this before. If you turn the multi-lang filter off, you should see that that feedback does indeed come out. It is just the way the multi-lang filter deals with spans that are immediately next to each other, leading to just the last one being outputted. By next to each other, I mean without a particular type of white space between the tags. Possibly newlines are different than spaces or some such thing. I think that people have been adding non-visible symbols to deal with this but I could not quickly figure out a simple solution.

We could probably make whatever extra is needed to be added between the feedback from the nodes, but that might affect the layout for some users.

It would be nice if the actual multi-lang filter were fixed, but while waiting for that, have you tried the `[[lang]]`-block? Basically, these are largely equivalent, since 4.4.2 + the fixes listed in #905:

```
<span class="multilang" lang="EN">...</span> = [[lang code="EN"]]...[[/lang]]
```
This gets processed entirely inside STACK and does not tie up any span-tags thus avoiding some annoying nesting issues.

alcarola:
Thanks a lot for your input! I by the way use STACK version 2022082900.

I tried your suggestion and discovered a bunch of problems; see the attached PDF (which uses the language setting Swedish (Finland):

a) The selected language is displayed only in the question text, but not in the general feedback (model solution) or PRT feedback. In the latter, the English text is displayed regardless of Moodle language. 

[...]

PR #1203, opened 2024-06-11
Update Languages.md

anst-i:
Added instructions how to avoid issue #940 to the Language documentation.

### DIFF
--- a/doc/en/Authoring/Languages.md
+++ b/doc/en/Authoring/Languages.md
@@ -110,6 +110,17 @@ But this question text causes no issues:
     <div>[[validation:ans1]]</div>
     <div>[[validation:ans2]]</div>
 
+### multilang in PRTs with several nodes
+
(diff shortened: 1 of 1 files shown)


sangwinc:
Thanks @anst-i that's really helpful.  We should probably improve the filter....

aharjula:
Well, someone should, not us though... But it seems like people are moving towards multilang2 from that original one. It just is not worth the effort to fight that filters parser, when it still would be finicky when working with things that may generate spans... 

And STACK itself moved away by building our own internal one so that it can be used in places that do not ever see the Moodle filters. Basically, if one wants to mix STACK-JS with localisation one cannot use any of those Moodle language filters inside those, only `[[lang code="XY"]]...[[/lang]]` works over there, Given that `[[jsxgraph]]`-content, is STACK-JS content, and often has some localised labels that really has some uses.

Naturally, it is annoying to have to use different syntax for this in different places, but if one really wants to use only one, I would go for multilang2. It works in most places within STACK but will not and never will work for STACK-JS containers.

anst-i:
I have noticed another bug in the multilang filter, so another solution would be welcome. It appears the multilang filter does not parse the content of the `<span lang=".." class="multilang">` tag properly and simply goes for the first `</span>` it can find, meaning you can't use `<span>` in multilang for styling.

PR #1218, opened 2024-07-08
Iss1217 - Failing tests in API on master

### LINKED ISSUE #1217: Failing unit tests in master from the API
I've just noticed that the master branch in Moodle now has some failing unit tests for the code in the API branch. It looks like something has changes in PHP
https://github.com/maths/moodle-qtype_stack/actions/runs/9810665112/job/27091276827#step:17:92

Could you take a look please @EJMFarrow?  Thanks

EJMFarrow:
Fixes #1217 

Various PHP update issues:
- Fix creation of MathJax URL using empty string rather than null.
- Add question description to API question loader to avoid testing against null.
- Rewrite API tests to avoid using a generic mock class which was causing a type mismatch issue.

### DIFF
--- a/vle_specific.php
+++ b/vle_specific.php
@@ -204,7 +204,7 @@ function stack_get_mathjax_url(): string {
             $urlstring = substr($mathjaxconfigurl, 0, $questionpos);
             parse_str($querystring, $queryparams);
             $queryparams = array_merge(['config' => 'TeX-AMS-MML_HTMLorMML'], $queryparams);
-            $querystring = http_build_query($queryparams, null, '&', PHP_QUERY_RFC3986);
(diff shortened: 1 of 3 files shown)


vle_specific.php
```diff
@@ -204,7 +204,7 @@ function stack_get_mathjax_url(): string {
             $urlstring = substr($mathjaxconfigurl, 0, $questionpos);
             parse_str($querystring, $queryparams);
             $queryparams = array_merge(['config' => 'TeX-AMS-MML_HTMLorMML'], $queryparams);
-            $querystring = http_build_query($queryparams, null, '&', PHP_QUERY_RFC3986);
```
sangwinc, line 207: Oh dear!  Thanks for sorting this out.

Issue #1212, opened 2024-07-05
Limit for systematically deployed variants


jonaslache:
Hello everyone,

In my opinion, the possibility to automatically deploy variants of STACK questions is a very valuable feature! However, I now have a question with a large number of variants to deploy. Using the approach described in the [docs](https://docs.stack-assessment.org/en/CAS/Systematic_deployment/), I wanted to deploy them all. Please consider an example with the following question variables where `x1` is the variable to be randomised:

```
list: makelist(k,k,1,150);
n1: mod(stack_seed-1, length(list))+1;
x1: list[n1];
```

Now, when I use the "Deploy seeds from one to:" feature within the dashboard with a value >100, I get the following error message:

![screenshot1](https://github.com/maths/moodle-qtype_stack/assets/131446471/2da07268-faf9-4939-8671-b0173d3c8731)

With entries from 1 to 100, it works fine, so I guess that there is a limit of 100 variants. I can understand that there might be runtime problems when a large number of variants are systematically deployed at the same time. But maybe we can have a feature like in the following mockup, where users can first deploy the seeds from 1 to 100, then from 101 to 200 and so on?

![screenshot2](https://github.com/maths/moodle-qtype_stack/assets/131446471/0af705ed-1c0b-42c1-847f-eeebc527f681)

I have attached my sample questionas a Moodle XML file.

Thanks and best regards,
Jonas

[sample-question.xml.zip](https://github.com/user-attachments/files/16105935/sample-question.xml.zip)

sangwinc:
Yes, we put in an arbitrary limit of 100 variants 
https://github.com/maths/moodle-qtype_stack/blob/master/deploy.php#L118

I just don't understand why anyone would need more than 100 variants of a particular question.  I appreciate that's my view (or lack of vision!) but we also have problems with performance when we have lots.  E.g. some colleagues were trying to deploy all variants with floats (literally 10^n variants) and that was causing serious problems.  So, we're going to have some limits and 100 was my choice for the systematic deployment option.

We could have the feature you suggest. It probably isn't that hard to implement, but it's not a high priority feature this summer.

jonaslache:
Thanks for your reply! I can certainly see that there is usually no need for >100 variants.

sangwinc:
OK, so I'm going to close this issue as I'm planning to take no further action.  If you'd like to add in the functionality as you proposed above ("Deploy seeds n to m") then please send me a pull request!

Thanks for your interest,
Chris

[...]

PR #1223, opened 2024-07-18
Added "Deploy seeds n to m" feature from issue #1212

jonaslache:
Dear Chris,

I have tried to implement my desired feature from Issue #1212 (Deploy seeds from n to m). This was my very first attempt with PHP. It would therefore be great if you could give me some feedback, also regarding naming conventions etc. 

Many thanks and best regards
Jonas


sangwinc:
Thanks @jonaslache that's much appreciated.  And you remembered to set up the pull request into dev.  I'll sort that out this morning and it will arrive in v4.7.0.

PR #1226, opened 2024-07-19
Default test

EJMFarrow:
Adds display of a default test to the question dashboard if a question has no tests.
Also runs the default test in the bulk test script if question has no tests.

(Not sure what the plan is with 4.7.0. We maybe want to leave this until after...)


sangwinc:
Thanks @EJMFarrow this is a very welcome contribution.

PR #1236, opened 2024-08-01
Response analysis refactoring

EJMFarrow:
First pass at tidying up the response analysis page:
- Separate calculation and display code.
- Divide calculation into functions.
- Create unit tests.
- Restrict to the responses from a single quiz. (Dropdown selection after initial page load.)
- Restrict to only use student responses.
- Each display section is now a tab.
- Create page using a mustache template rather than HTMLWriter.
- Collapse variants so only the top response is displayed initially.
- Visual tidying.

Basically the whole thing is the same as before just... nicer. This is a good point to stop for evaluation of what's been done so far and whether there's new functionality to add.


sangwinc:
Thanks @EJMFarrow that looks like a significant improvement.  Much appreciated.
Chris

PR #1238, opened 2024-08-02
Fix PHP deprecated errors and warnings

AnupamaSarjoshi:
Hi Chris, 

We have noticed the below PHP deprecated error and a warning in some of our questions. Could you please review the fix for the same?

Deprecated: preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in /var/www/html/moodle/question/type/stack/question.php on line 1572

Warning: foreach() argument must be of type array|object, null given in /var/www/html/moodle/question/type/stack/stack/questiontestresult.php on line 114

Thanks,
Anupama


sangwinc:
Thank you, that's much appreciated.
Chris

Issue #1240, opened 2024-08-02
issue with using Stack in Moodle 4.5 (latest development/main branch)


danmarsden:
require_once(/var/www/html/filter/tex/filter.php): Failed to open stream: No such file or directory in /var/www/html/question/type/stack/stack/mathsoutput/mathsoutputfilterbase.class.php on line 21

filter/tex/filter.php has been moved to: filter/tex/classes/text_filter.php in:
https://github.com/moodle/moodle/commit/29e7140e3029e3f94048619d4e3874b6c7fbe229

sangwinc:
@danmarsden Thanks for reporting this.  It popped up today in the continuous integration
https://github.com/maths/moodle-qtype_stack/actions/runs/10212640832/job/28256383588?pr=1238
and another user posted a query on the plugin page.   See below.

Clearly, since it is a failure of "require_once" it's catastrophic failure.  However, it's new and recent.  More than one person has reported it! We will investigate.

Chris

------------

Can anyone help me, I'm trying to update my STACK version to anything above 4.5 (I want to play with the new Geogebra options) I've tried using the installation process above and I'm getting:

There is a request to install plugin STACK (qtype_stack) version 2024072400 from the [Moodle plugins](https://moodle.org/plugins) directory on this site. However, this plugin is already installed on the site.

My current STACK version is 2023102700.

Current Moodle version is: Moodle 4.3+ (Build: 20231201)

Thanks in advance!

sangwinc:
Thanks @EJMFarrow for your very helpful email about this.  We think it's caused by a change in Moodle
https://tracker.moodle.org/browse/MDL-82427
Please note that the official release of STACK (https://moodle.org/plugins/qtype_stack) does not yet support Moodle 4.5.  Thanks for alerting us to this change in Moodle.  We'll sort this out for the next release of STACK (and any release which supports Moodle 4.5).

danmarsden:
@sangwinc I linked the commit from upstream moodle that caused this in the comment above which shows the cause and the moodle tracker issue (MDL-82427)

timhunt:
OK, so where we refere to the filter_tex class, we ar giong to have to cahnge that to something like

```php
if (class_exists()) {
} else {

}
```

Acutally, why am I writing code in a comment here. Let me send a pull request.

timhunt:
See pull #1242.

sangwinc:
Thanks @timhunt, I think we're now set for Moodle 4.5 on this issue at least.

PR #1247, opened 2024-08-12
Iss1240 - Movement of filter files by Moodle

EJMFarrow:
#1240 Here's another place where the file movement catches us, preventing TeX from appearing as an option.


sangwinc:
Thanks @EJMFarrow much appreciated.

PR #1250, opened 2024-08-16
Parsons history

smmercuri:
Add `log` option to the header of the `parsons` block. Setting `log = "true"` in the header results in the submission input containing the entire move history in an attempt along with the timestamp of that attempt (measured in seconds elapsed since 00:00 GMT 01/01/1970). The value is `log="false"` by default, which results in the submission input containing only the final state of the attempt. In order to integrate these two modes, the format of the submission input in the default `log="false"` case now takes the form 
```
[[{"used": [...], "available": [...]}], <timestamp>]
```

This PR also re-implements the legacy non-hashing versions of various authoring functions required for the parsons block, `proof_parsons_key_json` and `proof_parsons_interpret`, for compatibility. The hashed functions now take new names `parsons_encode`, `parsons_model_answer`, and `parsons_decode`.


sangwinc:
Thanks @smmercuri.  Adding in the hash was always in the plan (good tip from Sam) anyway.  The logs are potentially very useful for research.

PR #1260, opened 2024-09-06
Enhancement for factorlist(), multiple factors can be in the list

Mat-Ge:
Enhancement for factorlist() so that multiple factors can be in the list.
Are such enhancements to existing functions welcome? I am grateful for feedback.


sangwinc:
Thanks for this @Mat-Ge Yes, this kind of enhancement is very welcome indeed.  There is often some ambiguity about whether factors should be a "set" or a "bag".   Maxima doesn't have any support for bags (duplicates but no order), so a sorted list is probably a very sensible thing to have.  We really do need more discussion of the issues here, and the related issue of how to deal with repeated solutions to equations.  This is much appreciated. I'll check this out early next week and merge.

sangwinc:
Yes, your approach of retaining powers of factors is simple and direct and avoids the "bag" issue.  I'm happy with that solution. Thank you for not breaking back compatibility.

PR #1262, opened 2024-09-12
unit-test-fix - Skip certain asserts if Moodle < 4.1

EJMFarrow:
Skip 2 asserts in Moodle < 4.1 as format_text ignores noclean option in 4.0.


sangwinc:
Thanks @EJMFarrow much appreciated.

PR #1264, opened 2024-09-12
Parsons input

smmercuri:
This PR contains a new input type called Parsons that should be used for Parson's problems for proof and also matching problems using the `parsons` block. 

The main purpose of this input type is to facilitate readable state displays so that, while draggable items have hashed identifiers on the question page, the author sees the original non-hashed identifiers in submission data and on the test dashboard. 

Secondly, this input streamlines author experience when writing Parson's and matching problems since it is no longer necessary to use `<p style="display:none">` in Question text and `hideanswer` in the option field of the input.


sangwinc:
@smmercuri before I merge this all into dev, what will happen if people have old questions? In particular, what will happen with the changes in prooflib.mac if someone edits an older question and saves it?  That will load in the _new_ prooflib.mac.  Will the old question need changes? If so, we need a warning for any question which uses a parsons block and is written in an older version of STACK.  This is done here:  https://github.com/maths/moodle-qtype_stack/blob/master/question.php#L1508 

Does this make sense?

smmercuri:
> @smmercuri before I merge this all into dev, what will happen if people have old questions? In particular, what will happen with the changes in prooflib.mac if someone edits an older question and saves it? That will load in the _new_ prooflib.mac. Will the old question need changes? If so, we need a warning for any question which uses a parsons block and is written in an older version of STACK. This is done here: https://github.com/maths/moodle-qtype_stack/blob/master/question.php#L1508
> 
> Does this make sense?

The way this works is that all the old functions used when writing `parsons` questions, for example `proof_parsons_interpret`, are still in `prooflib.mac` and `matchlib.mac` in a subsection with `legacy` in the title. Old questions which use the `String` input type and these legacy functions will still work as they did previously, but keys will not be hashed -- I've just double checked this on my local site.

By the way, I've created some new labels for PRs `PR WIP` and `PR ready for review` to help signify the status of a PR ("still working on it"/"happy for it to be reviewed and merged" resp.). This one will be ready for review if the latest CI passes.

PR #1265, opened 2024-09-17
Iss1211 - Moodle App compatibility

EJMFarrow:
This is now ready for testing on Coulomb. (I don't have access/knowledge to set that up, though.)

I managed to re-write things so dropdowns can appear inline now. We have Behat test for all the normal input types and a basic JSX behat test. The CI setup changed for App plugins in 4.1 so these tests are only run against Moodle 4.1+, however.

GeoGebra might work but I can't test it locally due to mixed HTTP and HTTPS issues in my setup.

There are 2 changes to existing STACK code:

- Removal of login requirement for plots. Matti is OK with this.
- Small change to checkbox input to deal with the App sometimes returning odd data. I don't thing it will affect normal running of STACK but worth checking.


sangwinc:
Thanks @EJMFarrow.

PR #1267, opened 2024-09-24
Update Multiple_choice_questions.md to include partial marking

Bennett-Woolf:
This change adds to the documentation to show how partial marking can be done for MCQs


sangwinc:
Thanks, that's a very helpful addition to the docs.  Future contributions like this would be appreciated!

Issue #1249, opened 2024-08-15
Add test of size of embedded media to the bulk check system


sangwinc:
Colleagues are embedding large media into STACK questions, sometimes needlessly huge images.  This is a particular problem for students in low-resource environments who pay for internet by the MB.

We need a line in the bulk checker to alert when a question has "large" (tbc) images.  

This will also be a problem for version control, backup, etc.  There is just no need to put a 5MB image into a STACK question!

This should be a warning, not an error, just as we have warnings for "unused" media.

sangwinc:
A language tag for the error could go around here:

https://github.com/maths/moodle-qtype_stack/blob/master/lang/en/qtype_stack.php#L56

A size check (perhaps) could go here

https://github.com/maths/moodle-qtype_stack/blob/master/question.php#L1597

This is a "nice to have", so if checking the file size is really difficult then please just close this issue without resolving it for cost benefit.

PR #1271, opened 2024-10-01
Iss1249 - Add file size check in bulk test

EJMFarrow:
#1249 Downside of doing this here is that user-generated tests are not run. Are we in danger of masking a more pressing problem if users don't get round to pruning files?


sangwinc:
Many thanks.

Issue #1252, opened 2024-08-20
STACK breaks if 'forceclean' is enabled


geichelberger:
If you enable `forceclean` for your Moodle instance, STACK is not working correctly because the stack-iframe-holder div id gets stripped, breaking the `create_iframe` javascript function.

`Uncaught TypeError: document.getElementById(...) is null`
```js
create_iframe(iframeid, content, targetdivid, title, scrolling, evil) {
          const frm = document.createElement('iframe');
          frm.id = iframeid,
          frm.style.width = '100%',
          frm.style.height = '100%',
          frm.style.border = 0,
          !1 === scrolling ? (frm.scrolling = 'no', frm.style.overflow = 'hidden') : frm.scrolling = 'yes',
          frm.title = title,
          frm.referrerpolicy = 'no-referrer',
          evil ||
          (frm.sandbox = 'allow-scripts allow-downloads'),
          frm.srcdoc = content,
          document.getElementById(targetdivid).replaceChildren(frm),
          IFRAMES[iframeid] = frm
        }
```

The cleanup happens in `$question->format_text` before that line `$questiontext` contains the div with an id.

https://github.com/maths/moodle-qtype_stack/blob/ab90b2a6d22d6343934f8ee0f8e5d793caf79ba8/renderer.php#L56-L72

gergelyrakoczi:
This would be great to fix.

EJMFarrow:
In principle, adding the div id as a class and searching for that if the id is missing seems to fix things.
https://github.com/maths/moodle-qtype_stack/commit/444019e149c04b4ee2ccf03cce8f225324d1b42d
(Includes an extra check in input.js for #743)
Any thoughts, @aharjula ?

aharjula:
Well if it works... However, I do have to wonder what exactly is it with `id`-attributes that so troubles the cleaning logic? Are they somehow generally considered evil, or is this about cutting down scripting by making targeting by `id` difficult?

Would a data-attribute work or would it too be cleaned away? If so then maybe we could use e.g. `data-stack-iframe-id` instead so that we would not need to use `class` as that does have another role in the world in general. If that would work, then we could avoid having that fall-back in the logic.

EJMFarrow:
Thanks, @aharjula . Yes the cleaning does seem... severe... but I'll give the data attribute a go. If it doesn't get wiped, l'll check for that in our code but I'll leave the id on the element as well in case anyone else is relying on it.

[...]

Issue #743, opened 2021-09-28
Javascript error on quiz review page


georgekinnear:
I have a quiz that uses JSXGraph, and I noticed when reviewing an attempt that the graphs were not showing. My browser developer tools show this error:
![image](https://user-images.githubusercontent.com/30723394/135078072-6e953f47-f0d9-490d-a0dd-8d8567e343d2.png)
This seems to be the offending line of javascript code:
```javascript
    function initInputs(a, b, c, d) {
        for (var e = document.getElementById(a), f = !0, g = 0; g < d.length; g++) {
            f = initInput(e, b, c, d[g]) && f
        }
        if (f && (e.classList.contains("dfexplicitvaildate") || e.classList.contains("dfcbmexplicitvaildate"))) {
            e.querySelector(".im-controls input.submit").hidden = !0   /* this is causing the error*/
        }
    }
```
I couldn't locate this code in the codebase using search on GitHub, so I'm not really sure where this is coming from - but the mention of `dfexplicitvaildate` in this bit of code made me think it's STACK-specific.

I think it's because on the quiz review page, there are no check buttons being shown. I'm not seeing this behaviour when I review  a student's attempt using my teacher account, but it does arise when I use the "log in as" feature to view the quiz as if I'm the student.

(I realise this is not a super-helpful bug report, but hopefully it's meaningful enough for whoever wrote that code!)

aharjula:
Seems like a minified version of [this](https://github.com/maths/moodle-qtype_stack/blob/97b5f2fd7f7fe6481934d62fd9e73ac955bf17c8/amd/src/input.js#L431-L445), maybe that helps tracing it. This probably depends on the question themselves and the settings they have for inputs as well as possible scripting in the question affecting the input elements.

taraghi:
We observe the same problem while reviewing an attempt. We came across the problem as we tried to print the attempt on Chromium as PDF. The generated pdf is cropped. We believe that the Js errors are the cause.

Anyway the solution would be very simple here. One just need to modify the line https://github.com/maths/moodle-qtype_stack/blob/master/amd/src/input.js#L443 as follows:

```javascript
const input = questionDiv.querySelector('.im-controls input.submit');
if (input) input.hidden = true;
```

I hope that some one from Stack team cares of this issue in future releases

sangwinc:
@EJMFarrow could you review this please?  If still an issue could you propose a resolution?  Thanks,
Chris

EJMFarrow:
[...]

Issue #1144, opened 2024-03-20
STACK-JS switch_content produces body tags in content


bgailer:
Hello,

we have been using a lot of the new STACK-JS features recently and the functions to interact with the question content from inside the iframes are really helpful. One thing I noticed, when using the function `stack_js.switch_content` to change the content of an HTML element with an id like for example "target", then the content "hello" shows up enclosed by a HTML body tag.

For example, if I have
```
<span id="source">hello</span>
[[javascript]]
stack_js.get_content("source").then((content) => {
stack_js.switch_content("target", content);
});
[[/javascript]]
<span id="target">content should go here</span>
```
then the target span is
```
<span id="target"><body>hello</body></span>
```
which results in a different layout (block-level) and style of the target span than before. Instead, it should be just `<span id="target">hello</span>` in my opinion. Maybe this comes from parsing the content in the stackjsvle script.

aharjula:
Makes sense. This will be fixed to not use the whole `body` from the sanitiser parsing logic and only the contents.

aharjula:
Side note: the new `[[quid/]]`-block may be handy if you need to keep your identifiers unique:

```
<span id="[[quid id='source'/]]">hello</span>
[[javascript]]
stack_js.get_content("[[quid id='source'/]]").then((content) => {
stack_js.switch_content("[[quid id='target'/]]", content);
});
[[/javascript]]
<span id="[[quid id='target'/]]">content should go here</span>
```
Basically, if you find yourself copying the same pattern everywhere and have the risk of multiple questions on the same page using the same identifiers, then this might be useful.

georgekinnear:
I've got a side issue: the new `[[quid/]]` block sounds really useful, but it's not working for me. When I try to use it on the stack-demo site, I get an error message when saving the question: "Unknown block of type quid requested". Searching on github doesn't show up anything in the code defining `[[quid/]]`, so am I right in thinking this feature is still a work in progress?

aharjula:
It will come out with 4.6, there has not been a release since this was created.

It does [exist in the dev branch](https://github.com/maths/moodle-qtype_stack/blob/dev/stack/cas/castext2/blocks/quid.block.php), though... Simply dropping that single file into any 4.5+ level install should work.

georgekinnear:
Great, thanks!

PR #1272, opened 2024-10-01
Iss743 (and 1252) - Broken review on log in as + iframes broken by forceclean 

EJMFarrow:
#743 #1252 'Log in as' sets forceclean to true for security. Forceclean wipes the stack-iframe-holder id (and also any data-attributes so we can't use those instead).

I've added the id as a class as well so we can fall back to searching for that if the id isn't available. It feels like the issue is over-zealousness on forcecleans part but this gets us round it in a backwards compatible way.

Also fix another error on 'log in as' where JS was attempting to hide an element that wasn't there.


timhunt:
Here is a question:

Why are we calling Moodle format_text - which is what is doing the force cleaning - **after** we are generating other bits of HTML for widgts?

Normally format_text should only be called on the bits of user-input HTML.

(Sorry, I don't remember in detail how STACK works any more.)

aharjula:
Well as far as I understand it has always been called and thus will always be called, based on certain someones initial examples of how to build a renderer back in the early days of STACK3...

More seriously though I thought we are calling it so that the filters get applied, filters like multilang and MathJax and whatever else there is in play within the questiontext. Which is nice as then people can bolt whatever they want into the Moodle filter system and have it also working in STACK within that Moodle.

Is there an alternative way for calling just the filters but not the clean thing?

timhunt:
"Is there an alternative way for calling just the filters but not the clean thing?"

That is not the right qestion to ask. The question is

"How can we get it to apply to the bits of the question were we want filters appied, without applying to technical bits of HTML we are creating which we don't want messed up?"

aharjula:
Sounds like every CASText-block creator will be very happy. Basically, we need to carry with us from the individual block level the knowledge of which parts of the output are to be protected. We also must make it so that the question author cannot use that protection to skip "forceclean", basically, one must target it with utmost care in the blocks. This will mean that there will be new internal-blocks for those that need to signal this sensitivity during the compile phase and a change to the postprocess phase, the function signature for that will need to change to allow passing over information of these protected things.

Basically, we cannot know what sort of elements or syntax a given future block might need protected and we cannot be certain that these do not overlap, so the definition of the placeholders that will be in place during `format_text` will need to happen at the block level and those definitions will need to be passed around which will be fun but at least we then we will have a systematic way to skip this.

I'll do this next week and fix my blocks, but blocks that are not originally mine will need to decide on their own if they will do similar protections.

For those wishing to comment, this is the future:
 1. When postprocessing an evaluated bit of CASText one needs to give it a "placeholder"-object.
 2. Blocks can ask that object for unique placeholders to use in their output and register the original content into it.
 3. The return product from the postprocessing is now not directly usable as it might contain placeholders, but it should be filterable.
 4. After filtering one asks the "placeholder"-object to place things back.

Basically, every place that uses CASText will need to do that. There will be holders in various places for keeping track of those "placeholder"-objects.

aharjula:
https://github.com/maths/moodle-qtype_stack/pull/1272/commits/a1a3feff706f55e25316dfed9998bae66db5b5db should make the CASText system able to mark things it does not want to be filtered.

At least in my tests I managed to have a functioning IFRAME and a missing id from a manual element in the same question, running with "log in as". And that fix should automatically apply to all the blocks of the IFRAME-family.

The next issue is how do we build `id` elements in arbitrary elements we might want to target from STACK-JS? Do we need a new variant of `[[quid/]]` that would print out something more like the whole `id="...."` bit, but how would the placeholder for that stay in place, obviously the value of the `id`-attribute won't stay in place? Any ideas on how to do [this](https://github.com/maths/moodle-qtype_stack/issues/1144#issuecomment-2011626280) if the usage around the value coming from the system needs to be author sourced and thus needs to go though filters?

EJMFarrow:
Digging into what forceclean is actually doing, I think we can set the option `allowid` so the id doesn't get cleaned:
https://github.com/moodle/moodle/blob/0eacab846eebff126512f9738e26accef916e39b/lib/weblib.php#L1111
http://htmlpurifier.org/live/configdoc/plain.html#Attr.EnableID

aharjula:
Ok. So from the point of view of STACK-JS I really want that `allowid` option to be on, everywhere we filter CASText content. And I cannot see a reason why it would lower security to turn that on. Without it too many things become impossible, and there must be a way for a teacher to see what a particular student sees.

For larger generated HTML constructs that need to be secured and that come from CASText blocks we still need that placeholder logic, even if the current issues can be solved with just `allowid`. There might be blocks coming in the near future that will require larger HTML segment protection, and even if there were not we might now finally protect even against the HTML-tidy filter, though whether it lets `id`s through is another matter.

We will never protect anything author-generated that is not going to sandbox iframes from these filters. Anything other than those `id`-attributes that is.

EJMFarrow:
Yes, that sounds good. The stripping of the ids doesn't seem to be about security but about ensuring uniqueness and you've already dealt with that.

aharjula:
https://github.com/maths/moodle-qtype_stack/pull/1272/commits/d50548b14f52018cb0678b25d0b9508bdcd5c38b adds that `allowid` to that options and things do indeed seem to work.

To me, this seems ready, but for style points we need to decide:
 1. Will we revert that `class` attribute workaround?
 2. Do we want to override this [`format_text`](https://github.com/moodle/moodle/blob/67f5ee3cec6f1701bbbf1b6d57216a3d291e8802/question/type/questionbase.php#L440) or should we call the base implementation directly in our renderer?
 3. Is there anything else we might not want to clean? I, for one, don't think so, but maybe someone else has ideas?

EJMFarrow:
Hi @aharjula,

This looks good. I'll revert the class attribute workaround. I'd be happy sticking with overriding question->format_text as I think that makes it clearer what's going on. Can't think of anything else that needs protected.

PR #1275, opened 2024-10-04
Iss1253 - Library page improvements

EJMFarrow:
- Improve return navigation, so user is returned to a quiz if that's where they started from.
- Default category to quiz if that is where user started from.
- Add category selector.
- Basic display of non-STACK questions.
- Make inputs interactive so dropdown answers can be explored.
- Make it clearer when a question has been successfully imported.


sangwinc:
Thanks @EJMFarrow, this is really progressing well.  Much appreciated.

PR #1276, opened 2024-10-04
Update security-map.json

Mat-Ge:
Can we allow the variable "stringdisp" for teachers or are there any security or design concerns about this? 
I never use string functions or variables for calculations, but sometimes they could be useful for feedback. I often define a ternary function like _ternary(bool, exp1, exp2)_ anyway and it is sometimes clearer to use this function than an [[if]] [[else]] [[/if]] construction, e.g. in a singular/plural situation. Example: _{#n#} parameters are correct_ is grammatically wrong when n=1. So I like to write _{#n#} {#ternary(n=1,"parameter is", "parameters are")#} correct_, but then I get unwanted quotes in the feedback. There are surely many other ways to solve such problems, but an easy way would be to allow "stringdisp".

### DIFF
--- a/stack/cas/security-map.json
+++ b/stack/cas/security-map.json
@@ -7510,7 +7510,7 @@
       "function": "?"
   },
   "stringdisp": {
-      "variable": "f"
+      "variable": "t"
   },
   "stringout": {
       "function": "f",


aharjula:
Not ready to say if it would be bad or if there are some things that might not work inside a question having that toggled to something else.

But `{@"string"@}` is a thing. Do try if `{#n#} {@ternary(n=1,"parameter is", "parameters are")@} correct` would actually do what you want. Basically, if this works `{@if n=1 then "parameter is" else "parameters are"@}` a function returning a string should work also.

Mat-Ge:
Many thanks for your suggestion! I'm just noticing my two-month absence from Maxima and STACK with the confusing differences between {#, {@ in combination with Maxima strings and numbers. 
I will close this request.

PR #1278, opened 2024-10-09
Finalise Parson's input

smmercuri:
In this PR we polish the Parson's input. Namely Maxima CAS calls are embedded inside the input type to streamline functionality:
- The model answer is now input as `[ta, proof_steps]` compared to `parsons_answer(ta, proof_steps)`.
- If "hideanswer" is not used as an Extra option, then a proof display using proof block CSS styling is displayed automatically. Previously this required manual use of the `proof_display` function by the author.
- Update the documentation on authoring Parson's questions.


sangwinc:
Thanks @smmercuri, much appreciated.  This is really coming together now.

PR #1290, opened 2024-10-30
Fix CSS for Parson's proof headers

smmercuri:
The PR makes a minor fix to the CSS used in the `parsons` block, which makes the headers resize automatically in proof mode.


sangwinc:
Thanks @smmercuri.

Issue #1287, opened 2024-10-25
Question edit form can have clashing HTML id values


georgekinnear:
If you have a PRT with the same name as one of the inputs, the question editing form has non-unique id values in the HTML.

For instance, if you have an input called `ans1` _and_ a PRT called `ans1`, then the heading for both the input and the PRT have

`<fieldset class="clearfix collapsible collapsed" id="id_ans1header">...`

This is annoying because both sections then expand/collapse together.

I'm not familiar with how the editing form is put together, but from a bit of looking at the code I guess the issue could be resolved by replacing `header` with `input_header` and `prt_header` in the second argument of these lines:

https://github.com/maths/moodle-qtype_stack/blob/afcb387fefb428c4030ff8150e2ac9df288c2a8c/edit_stack_form.php#L423
https://github.com/maths/moodle-qtype_stack/blob/afcb387fefb428c4030ff8150e2ac9df288c2a8c/edit_stack_form.php#L532

(and also the other instance of `header` that comes up a few lines after each of these).

sangwinc:
My vote is that we require inputs and prts to have different names!

georgekinnear:
I don't see why that's needed, though. The only place it seems to be assumed is in the question editing form ID fields, at least from my experience of writing a few questions like this. Everything else works fine!

I found it useful in questions with many inputs that each have a corresponding PRT - e.g. example-generation questions where there are prompts for examples with different properties, where I might have `ans_polynomial`, `ans_nonneg`, etc. and just using the same names for the PRTs attached to each one made sense. (Arguably I could even have done without the `ans_` bit...)

timhunt:
> My vote is that we require inputs and prts to have different names!

Like others, I don't see the need, and think this would be a bad rule.

Just need to stick a prefix (I suggest prt_ or input_ on the field names when building the form.) I think ti is only the header elements, right, so nothing to do when the form is submitted.

sangwinc:
OK, bad idea on my part then!  We can fix the headers with a prefix.  @EJMFarrow please could you look into this?

EJMFarrow:
Headers updated for release as part of 4.8

PR #1293, opened 2024-11-04
Iss1287 - Differentiate headers in edit form

EJMFarrow:
Expand div IDs to differentiate sections in the edit form when opening and closing.

Fix CI while I'm here:
- iframes use a couple of sneaky class parameters to control what's going on. These can persist between tests and need reset  if their value is essential to the test.
- Behat for the Moodle App is no longer compatible with PHP 7.4. I've raised an issue and switched off our checks for Moodle 4.1 for now.
#1287


sangwinc:
Thank you @EJMFarrow that's great!  Thank you so much for fixing the headers in Parsons.
Chris

Issue #1325, opened 2024-12-03
4.6 - PHPUnit failures due to shared question banks


aspark21:
Testing Moodle 4.6 for https://tracker.moodle.org/browse/MDLQA-19535
STACK 4.8.1

These PHPUnit failures do not occur in the 4.5 version of the build with STACK 4.8.1 so must be caused by the Shared Question Banks changes in 4.6 even if the errors seem random

```
qtype_stack.editform_test/test_get_input_names_from_question_text_defaultqtype_stack\editform_test::test_get_input_names_from_question_text_default Attempt to read property "id" on bool  /var/www/html/lib/deprecatedlib.php:812 /var/www/html/question/type/stack/tests/editform_test.php:39 /var/www/html/question/type/stack/tests/editform_test.php:78 /var/www/html/question/type/stack/tests/editform_test.php:83 /var/www/html/lib/phpunit/classes/advanced_testcase.php:76

qtype_stack.editform_test/test_get_prt_names_from_question_defaultqtype_stack\editform_test::test_get_prt_names_from_question_default Attempt to read property "id" on bool  /var/www/html/lib/deprecatedlib.php:812 /var/www/html/question/type/stack/tests/editform_test.php:39 /var/www/html/question/type/stack/tests/editform_test.php:78 /var/www/html/question/type/stack/tests/editform_test.php:92 /var/www/html/lib/phpunit/classes/advanced_testcase.php:76

qtype_stack.library_import_test/test_no_webservice_accessqtype_stack\library_import_test::test_no_webservice_access Failed asserting that exception of type "core\exception\require_login_exception" matches expected exception "required_capability_exception". Message was: "Course or activity not accessible. (Activity is hidden)" at /var/www/html/lib/moodlelib.php:2626 /var/www/html/lib/external/classes/external_api.php:521 /var/www/html/question/type/stack/classes/library_import.php:88 /var/www/html/question/type/stack/tests/library_import_test.php:118 /var/www/html/lib/phpunit/classes/advanced_testcase.php:76 .  /var/www/html/lib/phpunit/classes/advanced_testcase.php:76

[...]

PR #1336, opened 2024-12-12
Fix CI issues for Moodle 5

EJMFarrow:
- Update Postgres.
- Deal with question categories only belonging to modules from 5.0 onwards.
- Handle some navigation changes.
- Make the set up for a role check test more specific to handle extra checking being done by Moodle.
- #1325


sangwinc:
Many thanks indeed.

PR #1338, opened 2024-12-17
Code tidy

EJMFarrow:
Deal with the updated coding standards.
The main issue is 1000s of missing or incorrect docblocks which would take months to add in a meaningful way. Instead, I've added instructions to ignore the specific code sniff at each affected point. This way we don't trigger the warnings/errors but we know exactly where we can improve the code if we're working on these areas in future.


sangwinc:
Thanks @EJMFarrow, this is much appreciated.
Chris

PR #1339, opened 2024-12-18
Healthcheck for AdminUI

Mat-Ge:
Was this intentional? The healthcheck was part of settings.php but not of the AdminUI.

### DIFF
--- a/adminui/index.php
+++ b/adminui/index.php
@@ -35,6 +35,8 @@ require_capability('qtype/stack:usediagnostictools', context_system::instance())
 $links = [
     get_string('stackDoc_docs_desc', 'qtype_stack',
             ['link' => (string) new moodle_url('/question/type/stack/doc/doc.php/')]),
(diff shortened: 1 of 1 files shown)


sangwinc:
@Mat-Ge this looks like an omission but the healthcheck isn't part of the "adminui" page.

We've exposed a lot of tools for _regular teachers_, including the bulk test.  But the healthcheck really is only for site admins.  regular teachers should not have access to this really (most won't know what it's for).

Thanks though, it does look like an error.

Mat-Ge:
@sangwinc, I see your objection that the Healthcheck is different from the other tools. Since all the other tools are intended for regular teachers rather than site admins, wouldn't it make more sense to reduce the link list in the STACK settings to two entries: AdminUI and Healthcheck? This would certainly be clearer than if the Healthcheck was somewhere between the other tools, which are basically intended for regular teachers.

PR #1343, opened 2024-12-19
Added title and alt tags to the string that is used in the editing form

marcusgreen:
I did not know what the icons meant as they are not right next to the dropdown in the PRT part of the form.


sangwinc:
Thanks @marcusgreen

PR #1372, opened 2025-01-24
WIP Iss1358 - Allow import of quizzes via STACK library

EJMFarrow:
Work in Progress but this is ready for some testing and feedback. Does is work as expected/required?

- Quickest way to test is to temporarily copy the `importtest` folder in the `samplequestions` folder into the `stacklibrary' folder and purge the MUC cache.

- Updated library structure display to not show Gitsync files, top directory and (where appropriate) default folders.
- Updated question render to show headings, intro and question file list from quiz files.
- Updated question import to handle Gitsync quiz files. Quiz is created, questions are imported and then added to quiz.
- Questions must be all from the quiz context but can be from multiple question categories.
- Quiz question category structure replicated on import.
- Added instructions to use Gitsync to add a course and its quizzes to the STACK library and maintain them.
- Added unit tests and a test quiz folder in sample questions.
- No behat test yet as we probably need a real quiz in the STACK library for that.
- Unit tests failing for main but looks like that's a PHPUnit change. We probably need to remove all versions of `assertMatchesRegularExpression` and `assertNotMatchesRegularExpression` in `test_base.php`.


sangwinc:
Thanks @EJMFarrow that's a huge amount of work, and a really valuable contribution.

Issue #1381, opened 2025-02-03
Parsons block needs better error trapping


sangwinc:
If you malform the teacher's answer then you get a PHP exception, and everything stops.  We should provide a better error to users.

Create a Parsons problem following Example 1 of https://docs.stack-assessment.org/en/Specialist_tools/Drag_and_drop/Parsons/

For the teacher's answer in the Parsons block don't use the (correct) `[ta, proof_steps]`, but instead use `parsons_answer(ta, proof_steps)` (plausible but wrong).

The question saves, but the dashboard cannot be loaded due to a PHP exception.

sangwinc:
At this point in the code
https://github.com/maths/moodle-qtype_stack/blob/master/stack/input/parsons/parsons.class.php#L43
we need to check we get something reasonable back from `json_decode` before we try to "count" it!

sangwinc:
@smmercuri this branch contains a starting point, but it needs much better error trapping, messages to the user and then some unit tests.

sangwinc:
Another user has reported this.  Here is the trace.

````
Exception - Cannot use object of type stdClass as array

[More information about this error](https://docs.moodle.org/401/en/error/moodle/generalexceptionmessage)

Debug info:
Error code: generalexceptionmessage × Dismiss this notification
Stack trace:
line 1046 of /question/type/stack/stack/utils.class.php: Error thrown
line 1061 of /question/type/stack/stack/utils.class.php: call to stack_utils::unhash_parsons_string()
line 167 of /question/type/stack/stack/input/parsons/parsons.class.php: call to stack_utils::unhash_parsons_string_maxima()
line 763 of /question/type/stack/question.php: call to stack_parsons_input->summarise_response()
line 702 of /question/behaviour/behaviourbase.php: call to qtype_stack_question->summarise_response()
line 109 of /question/behaviour/adaptive/behaviour.php: call to question_behaviour_with_save->summarise_save()
line 821 of /question/engine/questionattempt.php: call to qbehaviour_adaptive->summarise_action()
line 491 of /question/engine/renderer.php: call to question_attempt->summarise_action()
line 119 of /question/engine/renderer.php: call to core_question_renderer->response_history()
line 113 of /question/behaviour/behaviourbase.php: call to core_question_renderer->question()
line 907 of /question/engine/questionattempt.php: call to question_behaviour->render()
line 461 of /question/engine/questionusage.php: call to question_attempt->render()
[...]

PR #1388, opened 2025-02-11
Better error checking and optimisation for the Parson's input

smmercuri:
Addresses #1381

This PR contains the following:

- Improved error trapping and validation for the Parson's input
    - Add `validate_parsons_state`, `validate_parsons_string`, `validate_parsons_contents` methods to `stack_parsons_input` to check validity of a string containing a single Parson's state, a string containing a history of Parson's strings, and a contents array for the Parson's input respectively.
    - Add a new error message for the cases where the input does not pass the validation, and throw it in various locations in which the type and shape of the Parson's input is assumed in `stack_parsons_input`.
- Optimise performance of the Parson's input when used for logging history.
    - Add a new filter  `stack_ast_filter_909_parsons_get_final_submission` which simply returns the first element of the submission array (which corresponds to the newest state in the history). This is applied in the input base class only if the class name matches `stack_parsons_input` and ensures that only a single state is sent to Maxima for grading, rather than an entire history which can easily become too long and cause timeouts.
    - Reduce order of `stack_ast_filter_908_parsons_decode_state_for_display` from `909` to `908` so that it occurs before final submission extraction.
    - Update Maxima function libraries as now they expect to receive something of the shape `[{}, n]` vs. `[[{}, n]]`.
    - Auto-generate tests for the filters.
- Update unit tests for the Parson's input


sangwinc:
Thanks @smmercuri

PR #1398, opened 2025-02-15
Add Antidiff and AddConst answer tests

anst-i:
Provide ATAntidiff and ATAddConst calculus answer tests, with documentation


stack/maxima/sandbox.wxm
```diff
@@ -102,6 +102,37 @@ debug:true;
 /* [wxMaxima: input   end   ] */
 
 
+/* [wxMaxima: input   start ] */
```
sangwinc, line 105: Looks like you saved the sandbox in your development area!  Best to copy this somewhere outside the git repro to avoid committing working to the clean distro version.

sangwinc:
Thanks @anst-i that's a really helpful contribution.

stack/maxima/sandbox.wxm
anst-i: Thanks! And yes, that's a good idea.

Issue #1393, opened 2025-02-13
Analysis response page - results from quizzes where question selected at random are not visible


EJMFarrow:
The quiz dropdown does not work when a question is drawn randomly by the Moodle quiz. The quiz simply does not appear in the dropdown. 
- Well this is a pain but I think I've figured out a way of obtaining the correct quizzes from the info in the question_set_references table.

Also, button links contain `&amp;`s instead of `&`.
- The mustache template just needs extra curly brackets around the URLs.

EJMFarrow:
This is now fixed and will be part of STACK 4.9

PR #1406, opened 2025-02-21
Iss1393 - Issues with response analysis page

EJMFarrow:
#1393

- I've created a Behat test for the RA page in the hope of avoiding future problems like the jQuery issue.
- I've upped the number of repeats for Behat tests in the CI as this test and some of the App specific ones are struggling if GitHub is tired. (In this one, it's somehow possible to load the analysis page before the student's response has been recorded if things are running slowly.)
- URLs should be fixed.
- Including quizzes where the question is included as part of a random selection: This was gnarly. The random questions are stored as a filter meaning some involved computation is required to check each one for a given question. I've tried to restrict the number of checks as much as possible and there is some caching involved so performance should be fine in most cases. If someone had a lot of quizzes all taking random questions from a selection of subcategories of the same context as the question to be analysed, things might chug.

### DIFF
--- a/.github/workflows/moodle-ci.yml
+++ b/.github/workflows/moodle-ci.yml
@@ -182,5 +182,5 @@ jobs:
 
       - name: Behat features
         if: ${{ always() }}
-        run: moodle-plugin-ci behat --profile chrome
+        run: moodle-plugin-ci behat --profile chrome --auto-rerun 6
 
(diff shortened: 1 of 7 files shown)


sangwinc:
Thanks @EJMFarrow.

sangwinc:
@EJMFarrow can you please delete the branch and close the issue if we're done?

PR #1410, opened 2025-02-26
Bootstrap5

EJMFarrow:
- Fix collapsible sections in HTML broken by change to Bootstrap 5 in main.
- Fix unit tests broken by addition of an extra class to select tags in main.

### DIFF
--- a/templates/questionfolder.mustache
+++ b/templates/questionfolder.mustache
@@ -59,7 +59,7 @@
 
 <div class="qtype_stack_question_folder" style="margin-left: 30px">
   {{#isdirectory}}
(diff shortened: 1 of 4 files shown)


sangwinc:
Thanks @EJMFarrow this is really helpful.  I'm guessing that the CI failure is a glitch rather than a genuine failure. Accepting this PR will test that theory!

EJMFarrow:
@sangwinc The SBCL based maxima set up testing against Moodle 4.5 seems to be timing out as of today (without us having changed anything). I'm experimenting with that in another branch but now SourceForge is down. It's turning into a game of Whack-a-mole...

PR #1413, opened 2025-03-10
CI fix for Moodle 5

EJMFarrow:
Moodle has started rendering MathJax on the import page which messed up one of the behat tests as it was looking for the pre-render text.


sangwinc:
Thanks @EJMFarrow!

PR #1414, opened 2025-03-10
Api defaults basic

EJMFarrow:
- Add defaults for everything in the API question loader. It should now be possible for the API to fill in the blanks if a partial XML file is used. If there are no inputs, a single algebraic input will be created. If there are no PRTs a single AlgEquiv node will be created.
- Fix bug where API could not access the STACK library after previous code tidy.

### DIFF
--- a/api/public/cors.php
+++ b/api/public/cors.php
@@ -39,7 +39,10 @@ if (isset($_GET['question'])) {
 if (strpos($scriptname, '..') !== false
     || strpos($scriptname, '/') !== false
     || strpos($scriptname, '\\') !== false) {
-        die("No such script here.");
+        // Give a special exception for sample questions.
+        if (!($is_question && file_exists('../../samplequestions/' . $scriptname))) {
+            die("No such script here.");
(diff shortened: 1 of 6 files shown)


aharjula:
Just a slight snag... At some point, we had those STACK rendered information items, i.e. zero value (defaultgrade?), zero inputs and zero PRTs, just the question-text with all the handy CASText tools and CAS generated values. Can those survive when pushed through this logic? Or will they get extra inputs? Maybe if the question value has been set to zero, this creation logic could be disabled?

EJMFarrow:
Ah, yeah, they'll get an extra input and PRT. I'll take a look - thanks @aharjula !

sangwinc:
Oh dear!  I forgot completely about that @aharjula sorry.  @EJMFarrow this commit might help.
https://github.com/maths/moodle-qtype_stack/commit/70508637e00c2b065d00407af29ff511c1384af4
Please move the question elsewhere if there is a better place to put it.

Issue #1337, opened 2024-12-17
"Send general feedback to the CAS" tool does not generate new question version


jonaslache:
Hello everyone,

When creating STACK questions, I got used to the fact that all changes can be undone due to the question versioning feature introduced in Moodle 4. However, this does not apply to the ‘Send general feedback to the CAS’ tool, which can be accessed via the Question Dashboard: If you make changes to the question in this tool and click on ‘Save Back to question’, no new version of the question is created, i.e. the changes made in this way (possibly by mistake, because you actually wanted to click on the other button ‘Send to the CAS’) cannot be undone. Is it possible to improve the tool in this respect?

Thanks and best regards,
Jonas

sangwinc:
@jonaslache that was by design, rather than neglect.  But perhaps not expected.  We could change this, but typically I go through many versions of the worked solution and like it this way round!

jonaslache:
@sangwinc Thanks forn your reply, I actually didn't expect this behaviour to be intentional. I find it a bit counter-intuitive (in fact, exactly what I wrote happened to me: I accidentally clicked on the ‘’ button and then got frustrated that I couldn't undo the changes). But maybe that's just my opinion :-)

timhunt:
Just noting that maybe one day this Moodle core improvement will be made: https://tracker.moodle.org/browse/MDL-73701

Mat-Ge:
I'm a little late, but I would like to support Jonas and Andreas' opinion.
With "Save Back to question" it is not to be expected that the previous version will be permanently deleted. Since the introduction of versioning, a new version seems much more obvious to me and, above all, much more practical.

sangwinc:
OK, there seems to be general agreement we _should_ bump the version number.  That's an OK change because if you like the current behavior you just get new versions which can be tidied up later.

@EJMFarrow, could you please add this to your list and take a look?  Thanks.

EJMFarrow:
Given that each version is an entire new question, is there a reasonable way to achieve this without passing the changes back to the edit form and saving there, @timhunt ?

timhunt:
Worth considering the usability pros and cons of passing the data back to the edit form, and giving people a chance to confirm before saving the changes.

[...]

PR #1415, opened 2025-03-10
Iss1337 - Save as new version from CASchat

EJMFarrow:
#1337 

- Send the user to the edit page if they want to save their changes as a new version.
- Add tooltip and text to CASChat page.
- Add behat test.


sangwinc:
Thanks @EJMFarrow

Issue #1419, opened 2025-03-14
Database error when using "Find todo blocks" feature


jonaslache:
Hello everyone,

In our Moodle system, we are currently experiencing the following problem with STACK version 2025012100: When you click "Find [[todo]] blocks" in the question dashboard, a database error is triggered. Our Moodle admin has reviewed the logs of our PostgreSQL database. They indicate that the error occurs at the `REGEXP` point:  

https://github.com/maths/moodle-qtype_stack/blob/900d6453fc77e2980a148e340c6b84e6b4ea3a0d/stack/bulktester.class.php#L131C1-L137C40

According to our research, REGEXP seems to be a MariaDB syntax and is not compatible with PostgreSQL. The corresponding function in PostgreSQL, which in initial tests seems to produce neither errors nor results, is `~`. So `REGEXP` (MariaDB) = `~` (PostgreSQL). It would be great if someone could take a look at this!

Thanks and best regards,  
Jonas

aharjula:
I guess someone will need to build the query with [this](https://moodledev.io/docs/4.5/apis/core/dml#sql_regex) and test on some other databases.

PR #1423, opened 2025-03-18
iss1419 - Update todo search

EJMFarrow:
#1419 
- Update SQL to use Moodle LIKE wrapper.
- Add unit test for method (which means we've tested on mariadb and postgres).


sangwinc:
Thanks @EJMFarrow a testcase might have spotted that DB problem earlier of course.  My bad, as usual!

PR #1457, opened 2025-04-30
API wording

EJMFarrow:
- Add a blank footer file allowing pasting in of Edinburgh details.
- Redirect html/php requests to sample page.
- Add a link to the library page (stack.php) from the sample page.
- Fix PHP 8.3 issue with double declaration of a variable.
- Move copyright text on sample page to cope with short questions.
- Add fix for correct answers being tin the wrong order #1459 
Sample page


sangwinc:
Thanks Edmund.

PR #1458, opened 2025-04-30
WIP - Allow saving of a broken question

EJMFarrow:
- Adds an `isbroken` flag to STACK questions. This can be set using a tick box at the bottom of the edit page. It allows a STACK question to be saved even if STACK-specific validation fails.
- Save does not occur if the user needs to confirm removal of inputs/PRTs. A warning is displayed and the user must check the confirmation boxes.
- Save does not occur if there are Moodle validation errors. This will show a 'not saved' warning in most cases but not if the only problem is missing required fields - that's a browser-side check.
- Fixed existing bug with non-display of Moodle validation errors (e.g. set default mark to -1).
- Save does occur if the user goes straight for `Save` rather than `Verify and update form' (even when a new PRT is being added).
- Validation messages for a broken question are displayed on loading/reloading of the edit page.
- Requires a DB update for the new field.
- Broken questions throw a runtime error so are not displayed to students or on preview, etc.
- Restore, backup, export, import updated.
- Behat tests added.
- Added search for questions where latest version is broken to dependencies page.

This is at the stage where it could do with review. What have I missed?


EJMFarrow:
Updated to display the not-saved warning when validation fails normally (i.e. when the question is not marked as broken).

sangwinc:
Thanks @EJMFarrow

PR #1460, opened 2025-05-02
helm_parts styling should only apply to top-level list items

georgekinnear:
If you nest a list inside a "helm_parts" list, the styling is applied to the inner list by mistake:
image

This fixes the CSS so that the styling only applies to the list items at the top level:

image


sangwinc:
Thanks @georgekinnear that's a very helpful contribution!

PR #1467, opened 2025-05-09
Update docker-compose.yml

Mat-Ge:
I'm not a Docker expert, but with the current version of this YAML, I only got an error message. With my adjustments, the API works on my computer.

### DIFF
--- a/api/docker/docker-compose.yml
+++ b/api/docker/docker-compose.yml
@@ -1,7 +1,7 @@
 version: "4.0"
 services:
   maxima:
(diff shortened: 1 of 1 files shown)


EJMFarrow:
Hi, @Mat-Ge.

Thanks! Looks like I never created an image for the latest small version tweak. I'll do that today, hopefully, then the code in the repo should work as is.

EJMFarrow:
The correct STACK API image is now on Docker Hub so this change should no longer be needed. Let me know if there are any problems. Thanks for alerting me, @Mat-Ge !

PR #1470, opened 2025-05-13
Add static flag to API question render response

EJMFarrow:
https://github.com/IDEMSInternational/pretext/issues/3

Indicate if a question has interactive elements that make a printed version make no sense.

- Add `isinteractive` static variable to stack_cas_castext2_block.
- Updated on question compile if any blocks are interactive.
- Add flag to compiledcache.
- Pass flag as part of render response from API.

~~TODO - classification of body, escape, iframe blocks.~~

@geoo89 @sangwinc

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -75,6 +75,7 @@ The response is again a JSON document, with the following fields:
 - an int field `questionseed` indicating the seed used for this response
 - an int array `questionvariants` containing all variant seeds of the question
 - an array of arrays `iframes` of arguments to create iframes to hold JS panels e.g. JSXGraph, GeoGebra
+- a boolean field `isinteractive`, indicating if the question contains elements preventing a static representation. If true, a printed version of the question would make no sense
 
 The input configuration consists of the following fields:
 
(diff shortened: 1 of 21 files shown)


geoo89:
Thanks! Just did some minimal testing and it seems to be doing what it's supposed to.

I see there are also two [input types](https://github.com/IDEMSInternational/pretext/issues/2) "GeoGebra" and "Parsons", but I assume it's not possible to sensibly use them without the corresponding block? So no need to check for those I assume.

As for `<script>` (at least outside of iframes), that's is going to be deprecated at some point anyway, right?
Either way, easy enough to search for that manually if necessary.

sangwinc:
Thanks @EJMFarrow, @aharjula and @geoo89 for sorting this out.  Since all inputs are fundamentally "interactive" it doesn't make much sense for STACK to indicate these. I think, in the first instance, we should check on the API client side which inputs we can sensibly replace with a static alternative.  Do you agree @geoo89 ?

sangwinc:
Thanks for confirming. I think, then, I should accept this pull request and get the functionality into "dev" for use/testing/development.

PR #1472, opened 2025-05-14
update stack_publications to use biblatex and APA style

georgekinnear:


sangwinc:
Thanks @georgekinnear

PR #1475, opened 2025-05-15
Stack: Rename duplicate scenario names

AnupamaSarjoshi:
Hi Chris, 

When running Behat tests, our CI reports a warning about duplicate scenario names within the same feature file. And probably because of this it sometimes randomly fails on our CI server too. As each scenario in a feature file should have a unique name, I have renamed the scenarios. Could you please review?

Thanks,
Anupama


sangwinc:
Thanks, that's very helpful @AnupamaSarjoshi

Issue #1474, opened 2025-05-14
Setting "arccos(x)/arcosh(x)" as "Inverse trigonometric functions" option results in SQL error


anst-i:
Trying to set "arccos(x)/arcosh(x)" in the "Inverse trigonometric functions" option results in an SQL error, both when creating a new question or editing an existing one.

Error message:
```
[...]

PR #1478, opened 2025-05-20
iss1474 - Fix inverse trig option

EJMFarrow:
#1474 

- Change id string for inverse trig option from arccos-arcosh to arsinh so it fits in the DB field.
- Also change option on dropdown to arsinh(x) to match.


sangwinc:
Thanks @EJMFarrow.

PR #1491, opened 2025-06-06
Update for decimalplaces()

Mat-Ge:


Mat-Ge:
Thanks for merging @sangwinc .
You are probably right with your concerns. Would an initial ```if not numberp(ex) then return ex``` help?
More documentation is always welcome.

Issue #1485, opened 2025-06-04
Avoid characters in filenames breaking zip install


EJMFarrow:
We've had problems with certain characters in file names breaking install from zip file.

It seems Moodle's zip filehandler removes certain characters from file names and then (unsurprisingly) can't find the files later.

https://github.com/moodle/moodle/blob/c39b737063621f142fd8fb6bd05970acf6d4de73/lib/classes/param.php#L979

Characters we need to avoid are: ```&<>"`|':\/```

(GitSync currently avoids all these apart from ampersand and backtick. Although, maybe worth doublechecking escaped slashes - something funky goes on with them somewhere.)

sangwinc:
Thanks @EJMFarrow, much appreciated.

PR #1492, opened 2025-06-09
filename-edit - Remove apostrophes and ampersands

EJMFarrow:
Update Stack library to fix file names which break zip install #1485


sangwinc:
Thanks @EJMFarrow that's fewer than I feared!  Much appreciated.

Issue #1477, opened 2025-05-19
Should we include this to the local CORS-scripts collection of JS-libraries and CSS-styles.


aharjula:
So I was experimenting on how to transfer styles defined in stylesheets out from a sandbox `[[iframe]]` so that one could render stuff using random JavaScript tools inside such sandboxes and still bring the result onto the VLE-side so that the document is not fragmented by parts of it being displayed in iframes.

For reasons, I chose to experiment with the relatively common library [highlight.js](https://highlightjs.org/), and I think I got things somewhat working.

![Image](https://github.com/user-attachments/assets/9166143e-4544-426b-aa7e-423dedc3eea0)

I am not so sure if anyone needs to display code that often in STACK questions, but this could help; however, if we want to make it safe-exam-browser safe or usable in closed networks, quite a bit of code would need to be included in the STACK distribution within the CORS directory. Basically, a line would need to be drawn, and if this falls inside it, we need to then keep on upgrading one more thing, and I have not even checked if the licenses match.

If someone wants to play with this, it is available in the [iframecsstesting](https://github.com/maths/moodle-qtype_stack/tree/iframecsstesting)-branch.

jonaslache:
For me this would be a very useful feature. At my university, I am responsible for teaching lecturers to author STACK questions. This also includes the basics of JSXGraph, and in this context I have sample tasks in which a JSXGraph graphic can be seen on the right-hand side and the corresponding JS code on the left-hand side. However, I am currently also implementing it using the iframe block:

```
[[iframe width="100%" height="100%"]]
   [[style href="https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@11.9.0/build/styles/vs2015.css" /]]
   [[script src="https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@11.9.0/build/highlight.min.js" /]]
   [[script]]
       document.addEventListener("DOMContentLoaded", e=>{
          hljs.highlightAll()
        });
   [[/script]]

   <pre class="theme-vs2015"><code class="javascript">
var board = JXG.JSXGraph.initBoard(divid, {
   boundingbox: [-10, 10, 10, -10],
   axis:true
});

var image = board.create('image',['https://example.org/image.png',
   [-10,-10], [20,20]
],{
   fixed:true,
   highlight:false
});
   </code></pre>
[[/iframe]]
```

[...]

PR #1493, opened 2025-06-10
Iss1477 related features but no `[[highlightjs]]`

aharjula:
Some features and documentation related to #1477, but no `[[highlightjs]]`. The documentation shows an example of how to do the same as that block, though.

 - `[[entityescape]]` block for simpler code samples, no need to turn `<` -> `&lt;` by hand.
 - `[[style]]` block allows setting `crossorigin="anonymous"` which allows JS access to stylesheets in certain cases.
 - `stack_css_utils` is a very limited library that currently lets one inline stylesheet rules to elements, and thus transfer style from iframes without the stylesheets.


sangwinc:
Thanks @aharjula.  Sorry about the delay.  I was away last week and wanted to look at this quietly!

PR #1507, opened 2025-06-16
Authoring quickstart guide for JSXGraph questions

geoo89:


doc/en/Specialist_tools/JSXGraph/Author_quickstart.md
```diff
@@ -14,3 +14,194 @@ __We strongly recommend you do not use an HTML-aware editor when editing STACK q
...
+var line = board.create('line', [pO, pA], { fixed: true });
+var graph = board.create('functiongraph', '{#expr#}' , {name:'{#expr#}', withLabel: true});
+[[/jsxgraph]]
+```
+
+Note how we are using the student answer `ans1` in our plot, and we're drawing the graph of f(x) using a `functiongraph` object. Because our function is simple, its maxima string representation works as a definition for the `functiongraph`, but in general read [Basic_plots](Basic_plots.md) to see how to plot functions.
+
+### Blank inputs
+
+Note that if the student does not move the point, the input field will remain blank and they will not get any feedback. In our question, we made sure in our randomization that the initial position of A is not a correct answer, but this can be a trickier issue in particular with questions that have multiple points as inputs, as all of them may need to be moved for feedback to trigger.
```
aharjula, line 147: If you have a situation where feedback triggering depends on multiple bound inputs, and one might not actually move all those bound things. Try `stack_jxg.define_group([p1,p2,p3,...]);` after you have bound those points (or sliders), this will make it so that if any of that group moves, all of them are considered moved and the inputs will get values. Those moves do not propagate over groups, i.e., if there are groups `{p1,p2}` and `{p2,p3}` moving `p1` will lead to `p2` being considered moved but not `p3` however moving `p2` will mark both `p1` and `p3` as moved.

sangwinc: Thanks @geoo89 and @aharjula much appreciated!

PR #1518, opened 2025-06-25
Luke longworth fix scale nicely

LukeLongworth:


sangwinc:
Thanks @LukeLongworth

Issue #1512, opened 2025-06-23
MathJax 3.x and STACK in the (near) future


helour:
Will it ever be possible to use MathJax version 3 in the STACK? Moodle has no problem with it, but the STACK validaton field does not display math text.
![Image](https://github.com/user-attachments/assets/ac67a126-1d27-4036-80fc-202299b7e44b)
The top image is created with default MathJax 2.7.9.
The bottom image is created using MathJax 3.2.2.

EJMFarrow:
Hi @helour 

Thanks for raising this. It looks like a Moodle issue. Moodle 4.5 and below is counting on finding MathJax.Hub settings which are no longer part of MathJax in version 3:

https://github.com/moodle/moodle/blob/97ed6b7083cfb547c14ae3a2dcaf759079d48f66/filter/mathjaxloader/amd/src/loader.js#L74

This throws an error in the JavaScript before STACK has a chance to ask for the updated validation box to be reformatted.

Looks like Moodle 5 in fact requires an upgrade to MathJax 3. However, that does make think we should double-check how well MathJax is working with Moodle 5! (It seems OK so far.)

EJMFarrow:
Hopefully not too much, @sangwinc I've clearly been using MathJax 3 in main for a while and not noticed any showstoppers.  
- We should check what documentation we have about MathJax configuration (including conversion between versions) and update that.
- It's worth you checking your recent Moodle 5 install's MathJax settings and confirm it's MathJax 3 by default.
- Are there any questions worth checking with a careful eye on the MathJax display (e.g. ones with tables)? We might need to warn upgrading users in the docs.

helour:
I tried Moodle 5 with default MathJax https://cdn.jsdelivr.net/npm/mathjax@3.2.2/es5/tex-mml-chtml.js. So far I have only found a problem in Parson blocks.

![Image](https://github.com/user-attachments/assets/f5e66ee4-34e6-4acf-9a56-86993d284801)

[MathInductionParsonBlock.zip](https://github.com/user-attachments/files/20927829/MathInductionParsonBlock.zip)

smmercuri:
Ah yes, Parson's block needs to be told which MathJax version to use due to how they changed asynchronous typesetting. The code to work on MathJax 3 was already in the Parson's block, but there was no way to switch MathJax versions in the released code, so added that switch back in the above PR. In the next version of STACK you'll need to include `mathjax = "3"` in the block header:
```
[...]

PR #1523, opened 2025-07-03
Add MathJax 3 support to Parsons block

smmercuri:
In initial versions of the Parsons block we could switch MathJax versions between 2 or 3. This is required because Parsons block typesets LaTeX in draggable items asynchronously and MathJax does this differently between 2 and 3, so we require different code depending. The JavaScipt code to do so was already there, this PR introduces a function that parses the mathjax version from jsdelivr and cloudflare URLs, so that the Parsons block can adapt to the user config.  Mentioned in #1512


smmercuri:
Oh I see that `stack_get_mathjax_url` has changed since I wrote all this, I think it used to hard-code MathJax 2. Then yes I should update this accordingly

vle_specific.php
```diff
@@ -223,6 +223,36 @@ function stack_get_mathjax_url(): string {
...
+        }
+    }
+
+    if (preg_match('/(?:^|\.)cloudflare\.com$/i', $host)) {
+        if (preg_match('/mathjax\/(2\.\d+(?:\.\d+)*)/i', $url, $matches)) {
+            return $matches[1];
+        }
+    }
+
+    return null;
```
EJMFarrow, line 253: There are going to be people who have MathJax installed locally.  Can we fall back to checking for `mathjax.?2` or `mathjax.?3`? We should maybe then default to 2 (and up that to 3 in a year).

samwitzig: Thanks for mentioning this: Yes, please take locally hosted MathJax installations into consideration  - this is our setting on our Moodle installations. Thanks!

PR #1535, opened 2025-07-28
iss1519 - Add sticky footer

EJMFarrow:
iss1519

Put the 'Save and continue' and 'Preview' buttons in a sticky footer.

(Trying to add the 'Save' and 'Cancel' buttons as well proved tricky - I couldn't do it without copies of all four buttons in their current location as well as in the footer.)


sangwinc:
Thanks @EJMFarrow that looks really good to me.  I'll put this into dev now, and we will have the whole development cycle to trial this on various moodle versions.

Issue #1542, opened 2025-08-07
cli/bulktestall.php script when using the --id=?


sangwinc:
There is an issue with the cli/bulktestall.php script when using the --id=? option. 

When running this option we receive the following error message:
"Exception - Call to undefined method stack_bulk_tester::get_stack_questions()"

This has probably been broken since this commit
https://github.com/maths/moodle-qtype_stack/commit/4aaecc9ded96d1611bdf9d5c0f248f23dc234fef
fixing
https://github.com/maths/moodle-qtype_stack/issues/896

We should restore this functionality!

PR #1550, opened 2025-08-21
iss1542 - Fix bulktestall restart

EJMFarrow:
#1542


sangwinc:
Thanks @EJMFarrow that's very helpful.

Chris

Issue #1363, opened 2025-01-16
JSON structure for downloading students' response data


smmercuri:
Currently responses are recorded in the following format:
```
"Seed: n; ans1: ""..."" [score]; prt1: # = 1 | prt1-1-T"
```

To make parsing this more routing in offline analysis, creating an optional JSON equivalent of this would be beneficial. For example
```
{
    "Seed": "n",
    "inputs": {
        "ans1" : {
            "value" : "...",
            "status" : "score"
        },
        "ans2" : ... 
        },
    "prts" : {
        "prt1" : {
            "score" : "1",
            "node" : "prt1-1-T"
        },
        "prt2" : ...
        }
}
```

sangwinc:
@smmercuri, thanks for raising this.  I suggest you start with this method to create a JSON object.
https://github.com/maths/moodle-qtype_stack/blob/dev/question.php#L770
We can then decide how to give a teacher access to this method, e.g. adding functionality to the analysis script
https://github.com/maths/moodle-qtype_stack/blob/dev/questiontestreport.php

sangwinc:
@EJMFarrow I've just made one small addition, which is some minimal metadata about where this data comes from. In practical terms, we will need to record the source of data from multiple questions and we will need to know where this data comes from.  We might as well put this in the original data, in a consistent format, rather than have users add this themselves.  If they want more information about a question (e.g. how seeds map to evaluated question notes) then that's a completely separate issue!  This minimal information at least records which unique question, gives a human readable version (which may not be unique...), and the time.  It's tempting to add in server information, but I'm not doing that because it would be Moodle specific and then break ILIAS/API compatibility.  Users need to record the data source for themselves....

sangwinc:
@EJMFarrow I've just done an experiment with a quiz in adaptive mode.  

1st attempt at quiz:  answer was "x".
2nd attempt at quiz: answer was "x^2" (wrong), then another attempt with "x^3" (correct).

This is the JSON data:

`````````
{"id":"19609","name":"Count attempts","reporttime":"2025-08-28T14:18:47+01:00"}

{"inputs":{"ans1":{"status":"score","note":"","value":"x"}},"prts":{"prt1":{"score":0,"penalty":0.1,"note":["prt1-1-F"],"errors":[]}},"Seed":1}
[...]

PR #1553, opened 2025-08-25
WIP - JSON data on analysis response page - DO NOT MERGE

EJMFarrow:
#1363


EJMFarrow:
@sangwinc @georgekinnear Well, this got a bit gnarly... Turns out I was right the first time about the question steps and tries not entirely matching up (with the added bonus that the saved response summary is only for the most recent try). I've restricted the displayed steps to those with step data name '-submit' or '-finish' and calculated the response summary on the fly for any step that isn't the last.

I've tried this with adaptive, deferred, immediate and interactive question behaviours and it generally returns the records you would expect (although careful checking/testing of content of those records required). The anomaly is for interactive and adaptive where the student checks an answer and then submits the quiz - that answer returns two records, one with a 'todo' state and one with a graded state. This is not necessarily wrong since a student could submit a quiz without checking an answer but is it what we want? If further attempts are disallowed (e.g. no more hints), then submission does not produce another record.

I've also added more metadata to the JSON:
`{"userid":"6","qattemptid":"818","qstepid":"1490","qstepseq":"4","slot":"1","state":"gradedright","timecreated":"1756719416","inputs":{"R":{"status":"score","note":"","value":"(1\/9)*x^(-8\/9)"}},"prts":{"Result":{"score":1,"penalty":0,"note":["ATDiff_true.","Result-0-T"],"errors":[]}},"seed":2101103018}`

This still requires unit tests and probably some exception/null handling but I want to make sure the output is as you expect before pressing on.

EJMFarrow:
I'm now just retrieving the question attempts directly from the DB and using Moodle's submitted response iterator for the question attempt to determine the actual tries. This means we'll always match the Moodle response report in terms of recorded tries, irrespective of question behaviour.

I've also cut some of the JSON fields:
`{"userid":"6","qattemptid":"812","slot":"1","state":"gradedwrong","timecreated":"1756718689","inputs":{"R":{"status":"score","note":"","value":"1"}},"prts":{"Result":{"score":0,"penalty":0.1,"note":["ATDiff_var_SB_notSA.","Result-0-F","Result-1-F"],"errors":[]}},"seed":655340636}`

Not sure how useful the 'state' field is - that may depend on question behaviour.

sangwinc:
Thanks @EJMFarrow, JSON download of data is a really helpful addition especially for the researcher network (@georgekinnear ).

PR #1555, opened 2025-08-25
xmlize issue

EJMFarrow:
Update test to deal with deprecated and replaced method.


EJMFarrow:
FYI @sangwinc if you've got a working version of Moodle based on main then don't update it at the moment. The Moodle file structure changed big-time in July and the upgrade is not currently smooth...

Issue #1606, opened 2025-10-08
List questions flagged as "Broken" on the "TODO" page


sangwinc:
We now, very usefully, have the ability to save broken questions.

Please can we make it easier to find them in the DB?

(1) Please can we use the bulk test script to list any questions marked as "broken" in their own section of the report, just as we do with missing tests, no worked solution etc.  (We might do this already, I can't remember, sorry...).

(2) Please can we add a section to the /adminui/todo.php page which not only lists questions with an explicit "[[todo]]" block, but also broken questions?   Any question marked as broken has, implicitly, work "todo" I think!  Listing them all there as well would be very helpful.

Thanks,
Chris

PR #1615, opened 2025-10-27
Import validate

EJMFarrow:
This runs validation on imported XML questions.

- If issues are flagged, the question is marked as broken.
- Pointers to missing next nodes are set to stop.
- If the graph still can't be created, then fail import. (The edit form for such a question will throw an Exception so there's no point importing it.)
- Show issues and alterations on import page along with link to question dashboard if imported.

Also:
- Add questions marked as broken to 'Find TODOs' page. #1606 
- Separate validation warnings and file size checks from upgrade errors in bulk tests.
- Fix bulk tests to show all the issues raised, not just some of them.

TODO: importasversion needs fixed for us to show issues when importing using that (both in Moodle and via Gitsync).

### DIFF
--- a/questiontype.php
+++ b/questiontype.php
@@ -1592,6 +1592,14 @@ class qtype_stack extends question_type {
         $fromform->{$name . 'feedbackvariables'} = $format->getpath($xml,
                             ['#', 'feedbackvariables', 0, '#', 'text', 0, '#'], '', true);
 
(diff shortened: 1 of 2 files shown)


EJMFarrow:
@sangwinc This pull request should be fine to be checked and merged once the parser changes have been merged. It does not rely on importasversion.

PR #1635, opened 2025-12-12
Stack: Deprecated errors found in Stack bulk check

tailetan:
Deprecated errors found in Stack bulk check in many of the modules 

Deprecated: trim(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/html/moodle/question/type/stack/question.php on line 1769

image


sangwinc:
Thanks so much for your contribution @tailetan, much appreciate.  I've changed the base branch to stage this for the next official release.

PR #1669, opened 2026-02-05
Core compatibility with ILIAS: Added @ before some file functions to avoid warns

Saaweel:
Hi all,

**This PR adds the @ operator to selected file functions to suppress non-critical warnings.**

Regards,
Saúl from SURLABS.


sangwinc:
Thanks @Saaweel we'll sort this out somehow and make sure the core STACK has better support for ILIAS as needed.

PR #1671, opened 2026-02-05
Core compatibility with ILIAS: Use null as default value in lang param

Saaweel:
Hi all,

**This PR sets default values of certain language parameters to null, this value is considered one of the possible values, so we can deduce that, by default, it is null, so that it will not be necessary to call it null in all the places where it is used.**

Regards,
Saúl from SURLABS.


sangwinc:
Thanks @Saaweel that looks sensible, and it's something we'd certainly do if we needed to.

PR #1672, opened 2026-02-05
Core compatibility with ILIAS: Removed {} from js imports in blocks

Saaweel:
Hi all,

This PR removes unnecessary braces in JS, this is not a problem since, fortunately, all these modules export the same function that was being imported as the default value.

In this case, this change is important because in ILIAS, something like `{word}` without commas or spaces is taken as a template variable and, if it is not filled in automatically, it is deleted, breaking the functioning of the blocks. In other places where it is necessary to import with arms, fortunately this is not a problem, since, for example, `{one, two, three}` is not detected as a variable.

Regards,
Saúl from SURLABS.

### DIFF
--- a/stack/cas/castext2/blocks/adaptauto.block.php
+++ b/stack/cas/castext2/blocks/adaptauto.block.php
@@ -43,7 +43,7 @@ class stack_cas_castext2_adaptauto extends stack_cas_castext2_block {
         $list[] = new MP_String('script');
         $list[] = new MP_String(json_encode(['type' => 'module']));
 
-        $code = 'import {stack_js} from "' . stack_cors_link('stackjsiframe.min.js') . '";';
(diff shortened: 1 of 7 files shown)


sangwinc:
@EJMFarrow, @aharjula there are good reasons why the ILIAS folks need this change.  I'm not an expert in Javascript, so I need some advice before we accept this.

Should we accept this change (and follow that pattern in future) or should we add in a function to "locallib.php" to wrap library names in {...} for Moodle/Api/? leaving the ILIAS local lib to do no wrapping here?

aharjula:
@sangwinc , nothing to worry about here. As Saúl says, those are all imports of a default export and [those](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#default_import) do not need to have those {wrappers}.

Those were there just due to habit and copy-pasting, and should be safe to remove.

Issue #1521, opened 2025-07-02
Bulk test STACK questions in a particular quiz


sangwinc:
It would be useful if a teacher who can edit a quiz can bulk test the STACK questions in that quiz.

Related to #1498.

sangwinc:
We already have the "context" functionality in the bulk test, it's documented here: https://github.com/maths/moodle-qtype_stack/blob/dev/doc/en/STACK_question_admin/Bulk_testing.md

We could add question bank category, when we add in "quiz".  This just involved sorting out DB queries which return the list of question ids.

PR #1677, opened 2026-02-09
Iss1521 - Allow users to bulk test a quiz

EJMFarrow:
- Allow users with question:editall permission on a quiz context to run a bulk test on all the questions of a quiz.
- There's a new link from STACK dashboard to a page listing all the quizzes containing a question. Clicking a specific quiz runs the test for that quiz.
- Will test all the questions in random selections.
- It looks like there are lots of changes to bulktester class but it's mainly separating out the testing of an individual question into its own function. Comparing in VS Code gives a better impression of what's going on.

#1521

### DIFF
--- a/adminui/caschat.php
+++ b/adminui/caschat.php
@@ -226,7 +226,7 @@ if ($questionid) {
         $questionxmllink,
         stack_string('editxml'),
         ['class' => 'nav-link']
-        );
+    );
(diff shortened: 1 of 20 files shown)


sangwinc:
Thanks @EJMFarrow that's really very helpful. This makes it _so_ much easier to test the relevant questions.  Much appreciated.

PR #1709, opened 2026-03-17
NRW - External GitHub question libraries. WIP. Do not merge.

EJMFarrow:
- Allow access to GitHub question libraries from STACK library page.
- Allowed libraries are set on the STACK settings page via a JSON object.
- Cache of STACK directory structure and question renders for a GitHub library can be reset by any user with access to the library page. (They're much more likely to be changed than local libraries.)
- Involves much re-writing of the directory structure creation code in questionlibrary.class.php to make it more generic. (We'll need to carefully check the normal library has not be broken in any way.)

We may not actually want to contact GitHub directly but go through an NRW API. This is still useful groundwork, however.


EJMFarrow:
I had been thinking we'd merge the external question loads in two stages - Git and then NRW. (NRW builds on the Git changes.) It's been so long now, I'm not sure that will work. Going to close this and set up a pull request for the whole lot. Will keep both branches so we can still compare what's what.

PR #1719, opened 2026-04-07
Fix warning message about unused use statement.

vuvanhieu143:
Hi @sangwinc ,

There is a warning message repeatedly appearing in our error log:

Warning: The use statement with non-compound name 'category_form' has no effect in /var/www/html/public/question/type/stack/questionlibrary.php on line 36

I have removed that line, as it does not appear to be necessary.

Could you please review the PR? It is a minor change.

### DIFF
--- a/questionlibrary.php
+++ b/questionlibrary.php
@@ -33,8 +33,6 @@ require_once(__DIR__ . '/stack/utils.class.php');
 require_once(__DIR__ . '/stack/questionlibrary.class.php');
 require_once(__DIR__ . '/classes/form/category_form.php');
 
-use category_form;
-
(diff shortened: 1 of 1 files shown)


sangwinc:
Thanks for helping with this @vuvanhieu143, we'll take a look and sort this out!

sangwinc:
@EJMFarrow could you review this please as you added this line here
https://github.com/maths/moodle-qtype_stack/commit/b77610094dd86e418587941ad883ca1f52334d61

Thanks!

EJMFarrow (commented):
My bad. That's a good change. I notice I failed to namespace the category form which would be worth doing as well. I can do that as a separate pull request, though.

Issue #1081, opened 2023-12-05
Template blocks does not work in PRTs


jonaslache:
Hello everyone,

I defined a `[[template]]` block in the question text using `[[template name="foobar"]]...[[/template]]`.
When I then use the `[[template name="foobar"/]]` placeholder in the question text, it works great and the correct content is filled in.

However, it does not work when I use the `[[template name="foobar"/]]` placeholder in the PRT. Instead of the previously defined content, the message `Warning no template defined with name "foobar"` appears.

It would be great if the template blocks could also be used in the PRT text to avoid repetition.

Thanks and best regards,
Jonas

aharjula:
Well, this is a bit of a problem. Templates defined inside question variables (with inline CASText) should be available everywhere (as question variables are mostly available everywhere, validation being the messy exception), but those defined in question text are indeed only available in contexts where that question text is also being evaluated and PRT evaluation is not one of those contexts. Actually, there is only one evaluation context where it gets evaluated, and in that pretty much nothing else happens other than the initialisation of inputs.

To make this work, we would need to extract template content from the question text and transplant it into contexts where the question text is not present, and that is a bit more complex thing to do than it might seem like, at first glance. If mixed with localisation or other conditional definition it would quickly become a nightmare to extract.

Basically, it would make sense to define templates in the question text, but unfortunately, question text is a branch of the evaluation of the question happening so separately from everything else that transferring stuff from it to other branches is a problem. The current recommendation is to target the root of the "evaluation tree" if one wishes to share things around the question, and that root is question variables.

*Side note*, someone should really write a page of documentation on the way STACK questions are evaluated, where different parts happen, and which parts are visible to each other. In this case, the natural rule of what is higher up in the editor is also available lower down in the logic does apply as question text branches out, the same applies to general feedback and most notably while PRTs are evaluated in the same evaluation context they do not fully see each other and might not even be executed or present.

jonaslache:
Thanks, Matti, for commenting! Using the question variables instead of the template block actually is a workaround that solves my problem. Could it be an idea to add this information to the docs?

Just one question, before I open a separate issue for something that might be related and thus not to be solved easily: I recently noticed a similar behaviour for the `[[debug/]]` block: When placed in the question variables, it returns a list of all question variables and their values. But when I use it in a PRT, I don't get a list of question variables and/or feedback variables, but the following message: `["%cs","castext_debug_no_vars"]`. Is this related?

aharjula:
[...]

PR #1731, opened 2026-04-29
Update Static_blocks.md

georgekinnear:
Add note about defining templates in the question variables re: https://github.com/maths/moodle-qtype_stack/issues/1081


sangwinc:
Thanks @georgekinnear

PR #1733, opened 2026-04-30
Fix AlgEquiv fixture timeout

cameron1729:
## Summary
  This patch fixes a sporadic `answertest_general_fixtures_test` failure in the `AlgEquiv` fixture for  `258552*x^7*(81*x^8+1)^398` vs `x^3*(x^4+1)^399`.

The fixture expects raw mark `0` and an empty answer note, but on slower Maxima runs it can time out while `algebraic_equivalence` tries to factor `SA-SB`. When that timeout happens, STACK reports it as `ATAlgEquiv_STACKERROR_SAns`.

## Fix
Disable the expensive factor step by using `algebraic_equivalence_factorp:false`.

The expected result is unchanged: the expressions are still marked as not equivalent with no answer note.

## Precedent
Other `AlgEquiv` fixtures already disable specific expensive equivalence subroutines for known cases, including `algebraic_equivalence_trigexpandp:false` and an existing `algebraic_equivalence_factorp:false` fixture for large exponential expressions.


sangwinc:
Thanks very much @cameron1729 it's useful to have test cases of this feature, and know when it's needed.  Much appreciated.

cameron1729:
My pleasure @sangwinc - thanks for a speedy merge!

PR #1735, opened 2026-05-01
behat-ci-fix - Set Behat Selenium image

EJMFarrow:
Temporarily set Selenium Chrome version to an old version while we wait for a Moodle App update. Run App Behat tests against Moodle 4.2 and 5.2 (oldest and newest supported versions).


sangwinc:
Thanks @EJMFarrow that's a real help!

PR #1744, opened 2026-05-06
Update matrixfactorizations.mac

LukeLongworth:
Fix "diagonalize". It was not correctly orthogonalising the eigenspaces if the given matrix was symmetric as described.


sangwinc:
Thanks @LukeLongworth that's much appreciated.

PR #1745, opened 2026-05-06
Aug improvement

LukeLongworth:
Improvements to `aug` following suggestions from Marc Peterfi. The function behaves as it previously did when given a single matrix (that is, it will insert a vertical bar before the final column), but when given two or more matrices as arguments it will concatenate them with vertical bars between each. 

The original use case is to generate worked solutions for matrix inversion or similar, e.g. aug(A, ident(3)) → aug(rowop(A,2,1,A[2,1]/A[1,1]), rowop(ident(3),2,1,A[2,1]/A[1,1]) but could be used for more. 

aug(addcol(A,b)) and aug(A,b) are identical. 

Existing test cases will fail only because the vertical bar is now generated as a closing bracket instead of an opening one. This looks nicer, so I am treating this as a feature instead of a bug.

My apologies for two separate pull requests in quick succession!

### DIFF
--- a/doc/en/Topics/Linear_algebra/Matrix_library.md
+++ b/doc/en/Topics/Linear_algebra/Matrix_library.md
@@ -83,11 +83,15 @@ The last three predicates above will all accept an optional argument, `sp`, that
 
 It is sometimes convenient to be able to display an augmented matrix. `matrix.mac` adds some limited support for this. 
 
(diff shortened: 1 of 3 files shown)


sangwinc:
Thanks @LukeLongworth the `aug_matrix` function looks very useful.

PR #1747, opened 2026-05-06
Iss1680 - Tweaks to dashboard

EJMFarrow:
- Load latest version of question on page refresh even if in historic view (including telling browser to update dropdown selection, which it would rather cache).
- Display warning flags in tab headers even if no variants. (Code was just being skipped along with table sorting if no variants.) 
- Do not show test tab on call to deploy or test all. Test results for current variant no longer shown while waiting for variants tab to load if the user is deploying or is testing all variants. This is probably better?
- Add progress bar when deploying *and testing variants* from list or sequence.
- Don't show variant progress bar if no variants.


sangwinc:
Thanks @EJMFarrow I'm looking forward to using this all "in anger"!

PR #1751, opened 2026-05-08
Proof styles

georgekinnear:
This adds a new CSS class ".statement" that can be used to style the statement of a Theorem/Lemma/etc.


sangwinc:
Thanks @georgekinnear that's a very sensible addition.

Issue #1752, opened 2026-05-08
Inconsistent placeholders caused by mathjax filters


sangwinc:
The mathjax filter appears to be turning

```
\(\{\)
<span class="nolink">[[InMNEYMWDx:ans2]]</span><span class="nolink">[[GDBBdLBJLg:prt1]]</span>
\(\}\)
```
into
```
<span class="filter_mathjaxloader_equation"><span class="nolink">\(\{\)</span>
        &lt;span class="nolink"&gt;[[InMNEYMWDx:ans2]]&lt;/span&gt;&lt;span class="nolink"&gt;[[GDBBdLBJLg:prt1]]&lt;/span&gt;
        <span class="nolink">\(\}\)</span></span>
```
at this point: https://github.com/maths/moodle-qtype_stack/blob/master/renderer.php#L90

Note, since the `<span class="nolink">[[GDBBdLBJLg:prt1]]</span>` has been mutiltated to `&lt;span class="nolink"&gt;[[GDBBdLBJLg:prt1]]&lt;/span&gt;` the renderer.php throws an exception "Inconsistent placeholders".

I think this is caused by the un-matched opening `{`, but I'm finding this quite difficult to reproduce.

A minimal question reproducing this is attached.

[questions-Scratch-Placeholders-20260508-1601.xml](https://github.com/user-attachments/files/27528385/questions-Scratch-Placeholders-20260508-1601.xml)

sangwinc:
@georgekinnear, sorry your question seems to have thrown up this rather odd bug (email today about the analysis question).

sangwinc:
This is related to https://github.com/maths/moodle-qtype_stack/pull/1686

EJMFarrow:
Yeah, there definitely seems to be an issue with Moodle's Mathjax filter. It's searching for Latex maths in the text:
https://github.com/moodle/moodle/blob/168fb5d6a7491acf35d7d7a3deb555b19f4b3860/public/filter/mathjaxloader/classes/text_filter.php#L132
but then replacing '<' and '>' with HTML entities between curly braces ANYWHERE in the text:
https://github.com/moodle/moodle/blob/168fb5d6a7491acf35d7d7a3deb555b19f4b3860/public/filter/mathjaxloader/classes/text_filter.php#L255

So:
```
Type {@ta1@}
{[[input:ans1]][[validation:ans1]]}
```
will cause the problem.

[...]

PR #1753, opened 2026-05-18
iss1752 - Search for garbled nolinks when restoring placeholders #1752

EJMFarrow:
We can either not bother with the nolinks or search for the garbled versions like this (bearing in mind that in this curly brace enclosing the placeholder situation, the nolink will not function for filters which come after MathJax).
#1752


sangwinc:
I agree we should keep the "nolink" in (expecting this bug will be fixed...) and search for both as you suggest.  Thanks.

Issue #411, opened 2018-10-22
Drop load commands for complex i & j etc.


sangwinc:
Remove all files such as this
https://github.com/maths/moodle-qtype_stack/blob/master/stack/maxima/complexi.lisp
and replace internally with texput commands.

aharjula:
cos-1 vs. arches being one of these...

aharjula:
also multiplication signs.

sangwinc:
I'm on all of those Matti!

sangwinc:
Does anyone object to me dropping the mathml support as part of this issue?  See the above commit.

timhunt:
I don't have a problem with that.

PR #1765, opened 2026-05-30
Move contrib/chemistry.mac into core STACK Maxima function. 

### LINKED ISSUE #1764: Move chemistry.mac into the core
The contributed maxima chemistry.mac is large

https://github.com/maths/moodle-qtype_stack/blob/master/stack/maxima/contrib/chemistry.mac

To reduce latency in questions caused by loading this, we propose to move this functionality into the core.  This means the fixed content will be loaded directly by Maxima rather than physically copied into the question variables and hence subjected to the PHP parsing every time the question is used.

sangwinc:
Move contrib/chemistry.mac into core STACK Maxima function. Fixes #1764


sangwinc:
Comments on this would be welcome @aharjula, thanks.

stack/maxima/stackmaxima.mac
```diff
@@ -172,6 +172,8 @@ load("functs");
 /* Combined and processed maximasrc/ content. */
 load("maximasrccompiled.mac");
 
+stack_chemistry_declare(ex) := load("chemistry.mac");
```
aharjula: The thing with this that worries me is that this will trigger a load of a file at execution time. Even though most, if not all other things, are already inside the image, this would not be. It might be better to simply have the whole of "chemistry mac" inside this function.

If I recall correctly, some time ago, half a decade or more, there were cases where people had managed to generate a maxima image, but some logic failed because the image could not load external scripts related to some setting that loaded stuff in this style. That was probably due to issues with the Maxima execution environment, but I think we wrote open those loads to avoid that.

I think it was issue #411 where we removed various loads, in particular those Lisp loads, but I don't recall what the original cause for that was.

aharjula: I would do as follows:
 1. Write `chemistry.mac` as a definition of a function (`stack_chemistry_declare`) that defines all those values and things.
 2. And then top-level load that `chemistry.mac`.

This is to avoid having to go to the file system to load that separately, and instead have that bit already parsed in the image.

sangwinc: Thanks @aharjula I'd forgotten about #411, so thanks for this very helpful review!  Leave this with me.

PR #1767, opened 2026-06-01
Iss1680 - Dashboard deploy warnings

EJMFarrow:
Hide deploy messages on page refresh and add dismiss button for deploy warnings.


sangwinc:
Thanks @EJMFarrow that's great.

PR #1770, opened 2026-06-02
JSXgraph examples stack library

KZerva:
Added a folder with JSXGraph examples in the STACK library. I'm not sure if this is the right area to put them but there were already other JSXGraph examples there. Feel free to move these folders to a different area. I plan to add more examples in the future.


sangwinc:
Thanks @KZerva, I think that's a good location for these questions. Much appreciated.  I won't delete this branch, so you can carry on adding more examples here.

PR #1773, opened 2026-06-15
multiline - Nerf multiline comment checks.

EJMFarrow:


sangwinc:
Good solution @EJMFarrow, thanks.

PR #1775, opened 2026-06-18
Dashboard tweak - Add course and category display

EJMFarrow:


sangwinc:
Thanks @EJMFarrow that is actually quite a useful addition and one I'll take advantage of, especially when arriving at questions from, e.g. the bulk testing page.

PR #1776, opened 2026-06-19
replacebug - Check for null

EJMFarrow:
Entering e.g. `-2*ln(c(2e^-x-x))` for an input with insertstars 
```
    <input>
      <name>ans1</name>
      <type>algebraic</type>
      <tans>ans</tans>
      <boxsize>30</boxsize>
      <strictsyntax>1</strictsyntax>
      <insertstars>2</insertstars>
      <syntaxhint></syntaxhint>
      <syntaxattribute>0</syntaxattribute>
      <forbidwords>F,f,[,],{,}, In, IN, ln), log), log*, ln*, sin), cos), tan), sec),  csc), cot) ,sin*,cos*,tan*,sec*, csc*,cot*,[[BASIC-ALGEBRA]],[[BASIC-CALCULUS]</forbidwords>
      <allowwords></allowwords>
      <forbidfloat>1</forbidfloat>
      <requirelowestterms>0</requirelowestterms>
      <checkanswertype>1</checkanswertype>
      <mustverify>1</mustverify>
      <showvalidation>1</showvalidation>
      <options></options>
    </input>
```
was throwing:
`Exception - stack_maxima_lexer_token::set_end_position(): Argument #1 ($lastchar) must be of type stack_maxima_lexer_char, null given, called in [dirroot]/question/type/stack/stack/maximaparser/lexer.base.class.php on line 795`

I've initialized $last. Not sure if this is the correct approach or even if the initialization itself is correct, though. Should we be guarding against null where $last is used instead?

### DIFF
--- a/stack/maximaparser/decimal.comma.lexer.class.php
+++ b/stack/maximaparser/decimal.comma.lexer.class.php
@@ -264,12 +264,7 @@ class stack_maxima_lexer_decimal_comma extends stack_maxima_lexer_base {
     // phpcs:ignore moodle.Commenting.MissingDocblock.Function
     public function eat_number(stack_maxima_lexer_token $token): stack_maxima_lexer_token {
         $numbermode = 'pre-dot';
-        $last = new stack_maxima_lexer_char(
-            mb_substr((string)$token->value, -1),
-            $token->endline,
-            $token->endcolumn,
-            $token->endchar - 1
-        );
+        $last = null;
         if ($token->value === ',') {
             $numbermode = 'post-dot';
         }
@@ -287,7 +282,9 @@ class stack_maxima_lexer_decimal_comma extends stack_maxima_lexer_base {
             } else if ($c1->c === ',') {
                 if ($numbermode !== 'pre-dot') {
                     // No second dot nor dots in exponent.
-                    $token->set_end_position($last);
+                    if ($last !== null) {
+                        $token->set_end_position($last);
+                    }
                     $this->pushc($c1);
                     break;
                 } else {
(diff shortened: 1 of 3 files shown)


timhunt:
Thanks for looking at this issue so quickly.

I know very little about this, but the fix you have done does not seem to be consistent with the existing code.

As far as I can see, all the existing code tends to do 

```php
if ($last !== null) {
    $token->set_end_position($last);
}
```

So, probably best to keep things consistent.

timhunt:
In fact, given how many places have an if-statement like that, I wonder if it is worth adding a set_end_position_if_not_null(stack_maxima_lexer_char? $lastchar) method, to reduce the duplication?

And, might be good to review all calls to $token->set_end_position($last); ?

aharjula:
Ok. Back from Midsummer... Yet again, I managed not to drown. 

Sorry about the bug, not the first one where the logic for updating the token end position fails for a single char token. `if ($last !== null) {` as Tim points out is the pattern I would prefer, and I would avoid adding a `set_end_position_if_not_null` function, as in a lexer, we would like to avoid context switches as much as possible, on the other hand this is not a significant optimisation when we are dealing with this type of input.

Basically, some wrappings with those `if ($last !== null) {` for the exits of the exponent branch would be what I would choose. Basically, 788 and 795 are the ones I would think might cause issues.

EJMFarrow:
Thanks, both. I've updated with checks for null instead of initialisation.

timhunt:
The code looks better now. Who is going to do the final review and merge it.

aharjula (commented):
I am happy with this.

EJMFarrow:
Great. Thank you. Will pass over to @sangwinc

timhunt:
@sangwinc will you have time to look at this and merge this?

timhunt:
Thanks everyone for your speedy work here :-)

PR #1778, opened 2026-06-22
dashboard-table - Fix variant table for Moodle 5

EJMFarrow:


sangwinc:
Perfect @EJMFarrow, a single token fix in one file is always welcome!

PR #1782, opened 2026-06-25
Freetext fixes

EJMFarrow:
- Fix race condition when creating iframes in mobile app that was mainly a problem when testing locally but was sometimes causing iframe creation to fail normally. We now wait for all the iframe holders to be available not just the question div before creating the iframes.
- Always use 'config=TeX-MML-AM_CHTML' for MathJax in app iframes otherwise the setting was reverting when flipping between pages.
- Re-export the sample questions after saving and replace files. This seems to fix our line break issue within Moodle.
- In the API add carriage returns to the question variables to mimic saving in the Moodle edit form.


sangwinc:
Thanks @EJMFarrow

PR #1790, opened 2026-07-01
STACK: add line numbers/ highlight to Question and Feedback variables

vuvanhieu143:
Hi @sangwinc ,

As Chris Nelson mentioned in his email, we're looking to add line numbers and syntax highlighting for STACK questions.

I've implemented this using the Ace Editor library: https://ace.c9.io/#nav=higlighter

Since Ace doesn't provide syntax highlighting for STACK/Maxima out of the box, I wrote a custom highlighting mode for it. I'm not familiar with every aspect of the STACK language, so my implementation may not cover all cases. If you notice anything that could be improved or any syntax that isn't highlighted correctly, please let me know and I'll update the rules.

This is what the new UI looks like. At the moment, I've only enabled it for the **Question variables** and **Feedback variables** fields. The feature is enabled by default, but if we'd prefer to make it optional, I can add an admin setting to turn it on or off.

image

image


sangwinc:
Thanks @vuvanhieu143 that's really very helpful indeed.  This has been on our wishlist for sometime, and I'm very grateful to you for sorting this out.    For sure, we can improve the support for Maxima syntax.

I've changed the base to "dev" so this can be staged for the next release.  This is timely as we just released a new version (4.13.0) and are agreeing priorities for development for the next cycle.  This feature was on the list.  We'll review, test and I hope accept this.

Mat-Ge:
There is a VS Code extension that does a great job of highlighting Maxima's syntax. Perhaps some of this extension's highlighting rules could also be useful here for making Maxima easier to read. 
https://github.com/nicoewok/maxima-syntax-highlighting

amd/src/ace_editor.js
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * @returns {Promise<void>}
+ */
+const loadAce = () => {
+
+    if (acePromise) {
+        return acePromise;
+    }
+
+    acePromise = new Promise((resolve, reject) => {
```
EJMFarrow: Can we use requirejs rather than creating a script tag?

ace/ace.js
```diff
@@ -0,0 +1,23 @@
...{this.element.addEventListener("keydown",this.$onGutterKeyDown.bind(this)),this.element.addEventListener("focusout",this.$blurGutter.bind(this)),this.editor.on("mousewheel",this.$blurGutter.bind(this))},e.prototype.removeListener=function(){this.element.removeEventListener("keydown",this.$onGutterKeyDown.bind(this)),this.element.removeEventListener("focusout",this.$blurGutter.bind(this)),this.editor.off("mousewheel",this.$blurGutter.bind(this))},e.prototype.$onGutterKeyDown=function(e){if(this.annotationTooltip.isOpen){e.preventDefault(),e.keyCode===r.escape&&this.annotationTooltip.hide();return}if(e.target===this.element){if(e.keyCode!=r["enter"])return;e.preventDefault();var t=this.editor.getCursorPosition().row;this.editor.isRowVisible(t)||this.editor.scrollToLine(t,!0,!0),setTimeout(function(){var e=this.$rowToRowIndex(this.gutterLayer.$cursorCell.row),t=this.$findNearestFoldLaneWidget(e),n=this.$findNearestAnnotation(e);if(t===null&&n===null)return;var r=this.$findClosestNumber(t,n,e);if(r===t){this.activeLane="fold",this.activeRowIndex=t;if(this.$isCustomWidgetVisible(t)){this.$focusCustomWidget(this.activeRowIndex);return}this.$focusFoldWidget(this.activeRowIndex);return}thi
```
EJMFarrow, line 1: Do we need any of the files in the ace folder other than ace.js?

edit_stack_form.php
```diff
@@ -217,6 +218,13 @@ protected function definition() {
         if ($closebeforebuttonarr !== false) {
             unset($mform->defaultRenderer()->_stopFieldsetElements[$closebeforebuttonarr]);
         }
+        // Disable ace editor when running behat.
+        if (!defined('BEHAT_SITE_RUNNING')) {
```
EJMFarrow: This requires more explanation but it would be better if the editor was not disabled when running Behat.

EJMFarrow (commented):
Hi, @vuvanhieu143 . Thanks so much for this. Looks very useful. I think it needs a little tidying up before merging.

edit_stack_form.php
vuvanhieu143: Hi @EJMFarrow , The reason I disabled it for Behat is that the existing Behat tests can't be reused when the Ace Editor is rendered, unless we update the tests with custom Behat steps or use CSS-based workarounds.

Since we can't meaningfully test the Ace Editor itself with Behat, I don't think there's much value in enabling it during Behat runs.

Does that make sense?

vuvanhieu143:
hanks for the suggestions, @Mat-Ge. I've updated the Maxima syntax highlighting rules and added unit tests to cover the changes.

@EJMFarrow, I've switched the implementation to use RequireJS and removed the unrelated JavaScript files that were no longer needed. I've also explained why I chose not to render the Ace Editor when running Behat tests.

edit_stack_form.php
EJMFarrow: Hi @vuvanhieu143 thanks for fixing the other issues - that's great. The problem with disabling Ace for Behat is not to do with testing Ace but with with checking Ace (and any future change to it) does not break the edit form somehow on any supported version of Moodle. We really need to be testing the form with everything switched on.

It looks like updating createEditor in ace_editor.js to add `textarea.dataset.fieldtype = 'editor'` allows the Behat tests to find the correct input field and pass (in Moodle 5.2 at least).

vuvanhieu143: Thank you @EJMFarrow , I have updated the code, the behat is passed.

EJMFarrow: Fantastic. Thanks very much!

sangwinc:
Thanks so much @vuvanhieu143 that looks really nice!  It's been on my list of things to do for ages.  Much appreciated.

PR #1800, opened 2026-07-10
WIP: Backtick replace - do not merge

EJMFarrow:
Attempt to allow customisation of asciimath delimiters.

This is turning into more and more of rabbit hole. I've essentially got the use of custom delimiters working (or did at some point - I may have broken it a bit while going round in circles). This touches many places so we need to be sure what we want before proceeding.

The issue is deciding which delimiters to use. Do we:
- Always use the delimiters set in options.
- Use the delimiters that were set at the point the question was saved.
- Save the delimiters with the question, falling back to backticks if nothing present. If so, are these input or question level options? Trying to use the 'extra options' at input level is currently no good as there are parsing issues. Do we a DB change for this to work?


EJMFarrow:
We've decided this is not worth the changes involved.

PR #1802, opened 2026-07-13
Drop unused YAML PHP extension as optional dependency

ngandrass:
### Description

We are currently switching from the deprecated [PECL](https://pecl.php.net/) over to [PIE](https://github.com/php/pie). During this process we found that the STACK question type still has an optional dependency on the [yaml PHP extension](https://pecl.php.net/package/yaml).

The yaml extension was introduced in 2020 as per a57f18f6d0e68730383012349f8c1af3fa09f420 as a side-effect of requiring the mbstring PHP extension (probably an import from earlier?). 

The commit 3ceabf48264fdc1947fda5c59ed62dae04b87797 introduced the alternative implementation [symfony/yaml](https://packagist.org/packages/symfony/yaml) and 940bb616aecf07591819282b2f40a73871c36268 finally removed the last use of any yaml PHP extension function. Threfore I propose dropping the old yaml PECL extension from the list of optional dependencies.

---
### Changes

This PR:

1. Removes the PHP extension [yaml](https://pecl.php.net/package/yaml) from the list of optional PHP extensions as per abaf1a1e89750e65a076c3ecb44b69c10ddab966
2. Fixes the `environment.xml` by adding a missing `</PHP_EXTENSIONS>` closing tag as per fd30d7df5a3f6fd52e2b6036cdbb14661e18d254

### DIFF
--- a/api/docker/Dockerfile
+++ b/api/docker/Dockerfile
@@ -14,9 +14,7 @@ FROM php:8.4.5-apache-bookworm AS base
 #Install required php extensions
 RUN apt-get update && \
     apt-get -y install libicu-dev libyaml-dev libzip-dev zip && \
-    pecl install yaml && \
     docker-php-ext-configure intl && \
-    docker-php-ext-enable yaml && \
     docker-php-ext-install intl zip && \
     a2enmod rewrite && a2enmod actions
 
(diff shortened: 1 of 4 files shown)


EJMFarrow (approved):
Looks good, @sangwinc . Thanks, @ngandrass .

(API uses Symfony YAML via composer.)

sangwinc:
Many thanks @ngandrass.

PR #1804, opened 2026-07-14
nrw11 - Add more grading data in structured form

EJMFarrow:
- Add answer note in structured form from API grade and test routes.
- There's a change outside the API to facilitate this in stack/questiontestresult.php but this just adds more output from there so is hopefully harmless.


sangwinc:
Thanks @EJMFarrow.

PR #1805, opened 2026-07-14
Add line numbering to question variables and feedback variables.

sangwinc:


jonaslache:
Hello @vuvanhieu143 and @sangwinc,

I’ve just tested the new feature for code highlighting and line numbers, and I really like it! However, it seems you can no longer resize the "Question variables" field, is that right?

Before:
before

Now:
after

Best regards,
Jonas

vuvanhieu143:
Oh yes, @jonaslache, that should definitely be fixed. I was so focused on the functionality that I overlooked the usability of the UI. I will create another PR to fix it

PR #1810, opened 2026-07-17
STACK: Fix the js unit test failed due to missing ResizeObserver

vuvanhieu143:


sangwinc:
Thanks @vuvanhieu143 we'd noticed that!

PR #1814, opened 2026-07-17
behat-fix - Sticky footer fix

EJMFarrow:
- For some reason the sticky footer has got in the way of clicking the checkbox on main. It's not a real issue - I updated Moodle to check.
- Have worked round the problem in the test, setting the box without clicking.


sangwinc:
Thanks @EJMFarrow!

PR #1815, opened 2026-07-20
Xmlcompare - Allow comparison of question versions and/or XML files

EJMFarrow:
- Add new page accessible from STACK dashboard and AdminUI.
- From dashboard, allow comparison of question with previous versions. Allow upload of files for comparison with question or previous versions. Allow comparison of 2 files. Allow changing to another question (potentially non-STACK) in the same context. Edit capability on question required.
- From AdminUI, allow upload of 2 files for comparison. usediagnostictools capability required.
- Allow side-by-side and in-line comparison. Allow viewing of whole file or changes only.
- Edits shown at about a 'word' level, rather than individual characters as that was a mess. Result is hopefully a happy compromise of granularity, readability and code complexity. Might require some tweaking.


sangwinc:
Thanks @EJMFarrow that really is a very helpful feature indeed!

sangwinc:
I'm going to accept this pull request now, so we can start using it daily in the next development cycle, but keep this branch open in case we have further suggestions.

PR #1817, opened 2026-07-23
Freetext errors

EJMFarrow:
- Update ASCII extractors to return meaningful error messages.
- Use translations where available even though we're working on the front end.
- Display extractor errors to user in sticky container at bottom of ASCII block. Set `errors="false"` in parameters of extractor block to switch off.
- Fix use of 'string' rather than 'search' in laststringremainder extractor.
- Some general code tidying.
- Remove rather than update 2 tests from RTL code that were simply AI checking it hadn't reverted to its own previous mistakes and involved brittle string matching that these changes broke.


EJMFarrow:
@sangwinc I point us both to the PR description and included documentation: The errors need to be switched on for an extractor by setting `errors="true"` in the extractor block attributes. 😆

Example for 'Free-text input with multiple extractors' with errors for just the first extractor switched on.

Screenshot 2026-09-08 153225

sangwinc:
Thanks @EJMFarrow !

sangwinc:
Thanks Edmund,  after a little bit of cache chasing, that's working very well indeed.

1. Putting the option into each extractor is perfect.  I'd have probably put it in the ascii block, but this level of fine control is really valuable.
2. Please would you refactor this so that the default is errors=true?  We've been round a similar loop with feedback from answer tests.  Having to turn things _off_ means you at least know they exist in the first place!  

Otherwise, this is really exactly what I asked for.  Let's see how the extractors get used.

EJMFarrow:
@sangwinc I've updated so filters can now use the error display. It's a relatively small change in stackascii.js but I've also changed the extractorhelper file to a more generic asciihelper file and moved it so the change has touched all the extractors and their test files. 😞 I updated the calculation and cas filters to output errors. (NB Most of the error messages come directly from mathjs so won't be translated e.g. `{@ dfds @}` gives `Undefined symbol dfds`). I also discovered and fixed an existing bug where an incomplete operation was showing exception output in the main ascii display e.g. `{@ sqrt @}`

sangwinc:
Thanks @EJMFarrow that's great.  I think we are now done with this issue?

EJMFarrow:
Yes - it's just a case of trying it out in dev for now and tweaking as necessary.

sangwinc:
Thanks - the error message from the regex extractor is going to look very odd to students.  I suspect that's one people will want to turn off.  I suspect there might be a PR from an enthusiastic user about options for custom error messages....

Issue #1769, opened 2026-06-02
Adaptbutton in the PRT


StackcreatorEducation:
I would like to use the adaptbutton within the PRT. However, it only works when an additional question variable is generated; otherwise the adaptbutton does not function correctly.

**Without a generated variable:**

<img width="611" height="285" alt="Image" src="https://github.com/user-attachments/assets/49f63500-1e5c-401c-838d-d5982b8c7952" />

<br></br>

**With a generated variable: (the button works correctly)**
<img width="269" height="122" alt="Image" src="https://github.com/user-attachments/assets/9d856fe7-f454-46e7-a527-900e66d9ae22" />

<img width="516" height="126" alt="Image" src="https://github.com/user-attachments/assets/9e250074-a144-4259-a118-c8dbd9f9e52e" />
<img width="150" height="118" alt="Image" src="https://github.com/user-attachments/assets/d4d9f543-c864-4458-86c5-9c0e36a3f7b1" />

jonaslache:
I think @m-r-k is working on this problem. One workaround is to use the Moodle text format instead of the HTML text format:

<img width="1575" height="245" alt="Image" src="https://github.com/user-attachments/assets/38e7504b-a17a-43b4-8ae1-9b60b5fdd4de" />

aharjula:
I think this might be a #1085 style issue with flatness of the adaptbutton block. Basically, when we start seeing those lists in the output something has probably declared that it will always return a flat string when it does not actually do so and that causes issues when nothing else triggers more complicated output.

sangwinc:
@jonaslache are you able to run code in the "dev" branch?   We've merged in a change which I hope fixes this, but I've not had time to test if this really is the fix.

jonaslache:
@sangwinc Yes, I now tested it in the dev branch. The adaptbutton block now works in the PRT. I tried it using the plain text area editor with all four text formats (HTML, Moodle auto-format, Plain text and Markdown).

sangwinc:
Thanks @jonaslache much appreciated.

PR #1818, opened 2026-07-23
iss1769 - Change is_flat to false for adaptbutton #1769

EJMFarrow:
#1769


aharjula (approved):
Yes. Anything returning a list after processing is not going to be flat (that is, it always evaluates into a string in CAS) for optimisation, and this returns an iframe...

PR #1819, opened 2026-07-24
STACK: Fix Question variables field is overlapping with 'Save changes…

vuvanhieu143:
Hi @sangwinc 
I found a minor UI issue related to the CSS z-index, which was causing the Ace Editor to overlap with the sticky footer.

image

I've fixed the issue.

Could you please review the changes?

### DIFF
--- a/styles.css
+++ b/styles.css
@@ -1486,8 +1486,7 @@ USE OR OTHER DEALINGS IN THE SOFTWARE.
     color: #44331b;
 }
 
-#page-question-type-stack .stickyfooter,
-#page-question-type-stack .sticky-footer-content {
+[id^="page-question-type-stack"] .stickyfooter {
     z-index: 1050; /* Overrides Ace's z-index 4 & 6 */
 }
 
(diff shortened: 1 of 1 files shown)


styles.css
```diff
@@ -1485,6 +1485,12 @@ USE OR OTHER DEALINGS IN THE SOFTWARE.
 .que .comment {
     color: #44331b;
 }
+
+#page-question-type-stack .stickyfooter,
+#page-question-type-stack .sticky-footer-content {
```
EJMFarrow: Hi @vuvanhieu143 Thanks for fixing this. Much appreciated. I have a couple of questions:
- Do we really need to apply this to .sticky-footer-content as well as .stickyfooter?
- I'll look into adding Ace to the edit question XML page at some point. Can we make the change apply to ids beginning with 'page-question-type-stack' so it applies there as well?
Thanks!

EJMFarrow: Fantastic. Thank you!

vuvanhieu143: Thanks @EJMFarrow! Good points. I've removed the redundant .sticky-footer-content selector and updated the rule to use [id^="page-question-type-stack"] so it automatically covers all current and future STACK pages."

sangwinc:
Thanks @vuvanhieu143 much appreciated.

PR #1820, opened 2026-07-24
Iss1796 - Iframe scroll setting fix

### LINKED ISSUE #1796: Iframe block: Scrolling option not working
Hello,

when using the iframe block like this: `[[iframe width="100%" aspect-ratio="5/3" scrolling="no"]]...[[/iframe]]` it always generates the iframe on the HTML side using the attribute `scrolling="yes"` which is not expected.

I might be wrong, but it seems that the scrolling option is always a string when it is set, and thus it's always true in Line 217. 

https://github.com/maths/moodle-qtype_stack/blob/572c65643d6aa2d894a2c34f53ba5e8552a20eea/stack/cas/castext2/blocks/iframe.block.php#L175-L178

https://github.com/maths/moodle-qtype_stack/blob/572c65643d6aa2d894a2c34f53ba5e8552a20eea/stack/cas/castext2/blocks/iframe.block.php#L217

EJMFarrow:
#1796
Fix to convert 'false' (and strings other than 'true') to false when passed as manual scrolling parameter to bare iframe.

### DIFF
--- a/stack/cas/castext2/blocks/iframe.block.php
+++ b/stack/cas/castext2/blocks/iframe.block.php
@@ -174,7 +174,10 @@ class stack_cas_castext2_iframe extends stack_cas_castext2_block {
         }
         $scrolling = true;
         if (isset($parameters['scrolling'])) {
-            $scrolling = $parameters['scrolling'];
+            // ISS1796 - Fix to convert 'false' (and strings other than 'true') to false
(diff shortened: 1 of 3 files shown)


sangwinc:
Thanks @EJMFarrow that's great - I'll review this when I'm back.  Have a good weekend!

PR #1823, opened 2026-07-24
Fixing answer tests english docs list indentation

B28LH:
The [author tests English docs](https://docs.stack-assessment.org/en/Authoring/Answer_Tests/) currently number every test, rather than treating the examples under each group differently (e.g. in a sub list)

Screenshot 2026-07-24 at 6 13 47 pm

I believe switching for 2 to 4 spaces in the markdown should fix this.


sangwinc:
Thanks so much @B28LH that's very much appreciated Ben!

Issue #1824, opened 2026-07-25
Possible issue with hideanswer option after updating STACK


yositomi-opu:
I recently updated STACK to the latest version, and since then I have been unable to import question XML files that use the `hideanswer` option for answer variables.

More precisely, the XML files are imported, but they are marked as **broken**, and the questions do not work unless I remove the `hideanswer` option. Existing questions that were already installed seem to continue working for the moment, but this is a serious problem for us because we have a large question bank in which almost every question uses the `hideanswer` option. In addition, we can no longer import our existing XML files.

Has the `hideanswer` option been intentionally disabled for Radio buttons and Checkboxes? If so, could you let me know the recommended replacement?

If this is an unintended bug, I would greatly appreciate it if you could fix it.

Kentaro Yoshitomi

sangwinc:
Very sorry about this - it's almost certainly a bug.  We'll look into it and decide how to fix this.  If we need to make a "bug fix" release we will do that in this case.

EJMFarrow:
Sadly, it looks like the fix to #1720 broke things for hideanswer and allowempty in dropdowns and thus checkbox and radio inputs. This will affect 4.12 and 4.13.

I'm just running the CI on a minimal fix: https://github.com/maths/moodle-qtype_stack/compare/iss1824

sangwinc:
Thanks @EJMFarrow that's much appreciated.

sangwinc:
That's now fixed, and in the master branch.  We'll arrange an official "bug fix" release soon.

Issue #1720, opened 2026-04-10
Option "nonotanswered" ineffective in dropdown questions


wehro:
If I create a dropdown question with extra options ``hideanswer,nonotanswered`` the ``nonotanswered`` option is ignored. It works if I change the order of options.

sangwinc:
Sorry about that - should be fixed in the next release.  Thanks for raising it.

PR #1825, opened 2026-07-27
iss1824 - Fix hideanswer for dropdowns

EJMFarrow:
- Essentially this is a one line change with tests. Replaces the `break` removed by #1720 with a `continue` . This allows multiple options but prevents `hideanswer` and `allowempty` from hitting the `unknown option` error.

#1824


sangwinc:
Thanks @EJMFarrow

PR #1827, opened 2026-07-30
ci-update - Update postgres

EJMFarrow:
CI was failing on main because postgres version was too low.


sangwinc:
Thanks @EJMFarrow

PR #1831, opened 2026-08-03
Api frontend

EJMFarrow:
Adds an API Docker build where the browser does not have direct access to the API itself.
- Changes are self-contained apart from minor changes to stackshared.js
- Demo webserver has a page allowing search of STACK library question names and an example 'notes' page with teaching notes and interactive questions.
- A manifest of questions is built on Docker build, creating question IDs which are then used by browser requests and converted to XML on the webserver.
- Users can easily point to a different location within samplequestions and so use their own questions.


sangwinc:
Thanks @EJMFarrow that looks great to me.

PR #1832, opened 2026-08-03
Accessing GitHub libraries and the NRW database

EJMFarrow:
- Allow access to GitHub question libraries from STACK library page.
- Allowed libraries are set on the STACK settings page via a JSON object.
- Cache of STACK directory structure and question renders for a GitHub library can be reset by any user with access to the library page. (They're much more likely to be changed than local libraries.)
- Involves much re-writing of the directory structure creation code in questionlibrary.class.php to make it more generic. (We'll need to carefully check the normal library has not be broken in any way.)
- Allow search of NRW database, display of some retrieved metadata and import of questions.
- Allow upload of questions to NRW - page available from STACK dashboard. Returns NRW question id and whether the question is a duplicate.
- Separate capabilities for accessing external libraries and for uploading questions.
library settings
library search
nrw upload


EJMFarrow:
This has now been updated after the metadata merge and is ready to merge itself.

EJMFarrow:
@sangwinc I've checked this is still working and that the capability checks are on the webservice requests not just the pages. Should be fine to merge. Will bump the version number.

sangwinc:
Thanks @EJMFarrow

Issue #1830, opened 2026-08-03
stack_include does not appear to use Moodle proxy settings


yositomi-opu:
## Description

We are unable to use `stack_include` on our university's Moodle installation because access to external URLs times out.

Our Moodle server must use an HTTP proxy to access external websites. The proxy is configured in Moodle, and other Moodle functions can access external resources through it. However, the HTTP requests made by `stack_include` do not appear to fully respect Moodle's proxy settings.

This is not reported as an explicit Moodle or STACK error. Rather, the request to the external resource times out, and consequently `stack_include` cannot retrieve the requested content.

## Environment

* Moodle version: 4.5.4+ (20250530)
* STACK version: 4.9.1 for Moodle 4.0+ (2025040100)
* PHP version: 8.2.30
* Operating system: RHEL 8.10

## Relevant code

Our university's IT staff identified the following section of `vle_specific.php` as the likely source of the problem:

```php
if ($good) {
    if (!isset($cache[$translated])) {
        // Feel free to apply any proxying here if you want.
        // Just remember that $islocalfile might be true and you might do
        // something else then.
        if ($islocalfile) {
            $cache[$translated] = file_get_contents($translated);
        } else {
            $translated = clean_param($translated, PARAM_URL);
            $headers = get_headers($translated);
            if (strpos($headers[0], '404') === false) {
                $cache[$translated] = download_file_content($translated);
            } else {
                $cache[$translated] = false;
            }
        }
    }
    return $cache[$translated];
}
$cache[$translated] = false;
return false;
```

In particular, the call

```php
$headers = get_headers($translated);
```

appears to access the external URL directly, without using Moodle's configured HTTP proxy. In an environment where direct outbound HTTP access is not permitted, this request times out before the content can be retrieved.

The comment in this section also seems to indicate that proxy handling has not yet been implemented here:

```php
// Feel free to apply any proxying here if you want.
```

## Expected behaviour

All HTTP requests made by `stack_include`, including the request used to check the response headers, should respect Moodle's standard proxy configuration.

## Actual behaviour

The request to the external URL times out because it appears to bypass the HTTP proxy configured in Moodle. As a result, `stack_include` cannot be used on our Moodle installation.

## Request

Would it be possible to replace the direct HTTP access in this section with Moodle's standard HTTP client or another method that respects Moodle's proxy settings?

Please let us know if any additional configuration details or diagnostic information would be helpful.

EJMFarrow:
[...]

PR #1834, opened 2026-08-04
iss1830 - Remove superfluous header checks.

EJMFarrow:
#1830 - Remove header check for 404 that was not using proxies and rely on `download_file_content()` to return false if the file can't be obtained for any reason.

### DIFF
--- a/vle_specific.php
+++ b/vle_specific.php
@@ -348,19 +348,14 @@ function stack_fetch_included_content(string $url) {
 
     if ($good) {
         if (!isset($cache[$translated])) {
-            // Feel free to apply any proxying here if you want.
-            // Just remember that $islocalfile might be true and you might do
-            // something else then.
             if ($islocalfile) {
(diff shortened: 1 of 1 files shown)


aharjula (approved):
Should be fine. Even if `download_file_content` returned the body from a 404 response, things would still break anyway; unfortunately for CASText, it would include it as nonsensical content which should be visible, but CAS includes would very likely not be parseable and thus explode.

sangwinc:
Thanks @EJMFarrow

yositomi-opu:
Thank you!!

sangwinc:
@yositomi-opu can you use this fix, or do you need a new official release?  (Some people can only install an official release).

yositomi-opu:
Thanks, Chris. I think we would need an official release in order to apply the fix to our university’s Moodle installation. Could you please include the fix in the next official release?

PR #1835, opened 2026-08-05
Add line numbers to XML edit

EJMFarrow:
- Add the ace editor to the XML edit page.
- Involves a little bit of meddling with the ace set up as it doesn't allow setting of the number of rows, only actual height, so we need to base height on the height of the original text area. (Sigh.)


sangwinc:
Thanks @EJMFarrow I was also wondering about this, and also about adding ACE to the new free-text input!  Is there a markdown option?

PR #1843, opened 2026-08-10
Core compatibility with ILIAS: Add helpers for platform-specific paths and input rendering

Saaweel:
Hi all,

As part of the ongoing STACKForILIAS 11.0.0 work, this PR adds a few small extension points to STACK core that allow ILIAS to reuse the same code without carrying local copies of Moodle-specific paths and rendering details.

The changes are intentionally minimal and preserve existing Moodle behaviour:

- Add `stack_plot_url_base()` and `stack_cors_scripts_dir()` in `locallib.php`.
- Use these helpers when rendering question-library plots and checking JSXGraph/Parsons styles.
- Extend `replace_validation_tags()` with an optional custom validation value.
- Add `get_name()` to the base input API.
- Keep equivalence validation markup unchanged in Moodle, while using valid block markup outside Moodle.
- Purge cached CAS results when their STACK-Maxima version is incompatible.

These changes are needed by STACKForILIAS, but are neutral for Moodle and reduce future divergence between both integrations.

Validated with PHP lint and on Moodle 5.2 using the STACK healthcheck, including CAS, cache, plots, JSXGraph and Parsons.

Kind regards,
Saúl
SURLABS


sangwinc:
Thanks @Saaweel that's really very helpful.  We'll review these (when the holidays are over) and sort out some mechanism to implement the changes you need.  Please let us know of anything similar.

stack/input/equiv/equiv.class.php
```diff
@@ -620,7 +620,7 @@ public function render_validation(stack_input_state $state, $fieldname, $lang =
 
         if (self::INVALID == $state->status) {
             $feedback .= html_writer::tag(
-                'span',
+                defined('MOODLE_INTERNAL') ? 'span' : 'div',
```
EJMFarrow, line 623: It would be worth adding a comment here so this doesn't get 'tidied up' by a Moodle developer in the future.

EJMFarrow (commented):
Looks great. Just one comment required as far as I can see.

stack/input/equiv/equiv.class.php
Saaweel: Thanks, I agree with you. I've added a comment to clarify why the markup differs between Moodle and ILIAS.

stack/input/equiv/equiv.class.php
```diff
@@ -620,6 +620,8 @@ public function render_validation(stack_input_state $state, $fieldname, $lang =
 
         if (self::INVALID == $state->status) {
             $feedback .= html_writer::tag(
+                // Moodle renders validation feedback inline, so preserve the span markup used there.
+                // ILIAS requires a block element for its alert styling to render correctly.
```
sangwinc, line 624: Thanks @Saaweel

EJMFarrow (approved):

sangwinc:
Thanks @EJMFarrow. I'll merge this into dev.

PR #1851, opened 2026-08-25
Matrix inputs: expose delimiter shapes to assistive technology

adamant-pwn:
### Summary

Following the accessibility discussion in #1849, this exposes the visual delimiters around fixed and variable-size matrix inputs to assistive technology.

Each visible left and right delimiter has a corresponding empty `span` with `role="math"` and a localized `aria-label`. This covers square brackets, parentheses, curly brackets and vertical bars. The no-brackets setting emits no accessibility elements.

The existing visual CSS and input API metadata are unchanged.

### Verification

- Added data-driven renderer tests covering all five `matrixparens` values for both matrix input types.
- Updated existing exact-HTML renderer tests for the additional elements.
- Rendered fixed and variable-size inputs through the standalone STACK API.
- Verified in a browser accessibility snapshot that the order is left delimiter, input, right delimiter.
- Verified that the empty elements have zero width and do not change the variable-input dimensions.
- Rebasing onto the current `iss1848` branch produced no conflicts, and PHP syntax checks pass for all modified PHP files.

### DIFF
--- a/doc/en/Authoring/Inputs/Matrix_input.md
+++ b/doc/en/Authoring/Inputs/Matrix_input.md
@@ -18,7 +18,8 @@ We cannot use the `EMPTYANSWER` tag for the teacher's answer with the matrix inp
 
     ta:transpose(matrix([null,null,null]));
 
-The brackets around the HTML input grid are taken from the question-level `matrixparens` option.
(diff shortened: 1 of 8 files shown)


sangwinc:
Thanks @adamant-pwn that's very helpful and much appreciated.

Issue #1848, opened 2026-08-22
Provide separate bracket-style hooks for matrices and row/column vectors


adamant-pwn:
Following #1847, a fixed grid can preserve whether its model answer is `matrix(...)`, `c(...)`, or
`r(...)` and expose that type to HTML/API consumers. Could STACK use that distinction to provide
supported bracket-style hooks for general matrices, column vectors and row vectors?

There are currently two separate rendering paths:

- CAS TeX: `matrix(...)` uses `stack_matrix_disp()` and `lmxchar`, while
  `stack_linear_algebra_declare(true)` hardcodes `bmatrix` for both `c(...)` and `r(...)`;
- fixed-grid HTML: every value type follows the one question-wide `matrixparens` option.

A question containing both semantic kinds therefore cannot choose their notation independently across
question text, validation/feedback and the HTML input grid. For example, square matrices and round
column vectors currently require custom `texput` definitions, author-added HTML wrappers and a copied
CSS renderer.

The desired API would:

- preserve current behavior as the compatibility default;
- use the model-answer value type rather than infer semantics from dimensions;
- allow `matrix`, `c` and `r` delimiters to be selected independently on the CAS side; and
- expose corresponding square, round, bar and no-bracket hooks for the HTML input and API.

This is not a request for STACK to inspect Moodle course context. It is a request for the HTML input
widget and CAS display code to expose the same semantic distinction already present in the value. The
exact option, Maxima-variable or CSS API is open for discussion. Historical context: #109 introduced the
`c(...)`/`r(...)` convention, #526 explains why the grid is HTML/CSS rather than MathJax, and #645
records the resulting separation between `lmxchar` and input-grid brackets.

If this direction is acceptable, I would be happy to prepare a focused follow-up PR after #1847.

sangwinc:
Thanks @adamant-pwn for looking into this so carefully and thoroughly.  That's very much appreciated.

Yes, we missed making `c` and `r` respect the options for parentheses types and hardcoded `bmatrix` for both `c(...)` and `r(...)`.  That's clearly not right, and needs to be fixed.

I agree any changes to inputs should use the model-answer value type rather than infer semantics from dimensions.

Nobody has made a compelling case to allow `matrix`, `c` and `r` delimiters to be selected independently.  My view was that `c` and `r` are just sugar for `matrix`.  That said, it's not for me to dictate things like this.  If we need this option, then we should support it!

@LukeLongworth do you have any comments here as well?

[...]

Issue #1327, opened 2024-12-03
Feature Request: "Context-aware" Block


LukeLongworth:
Kia ora, 

This is possibly a long shot, but I thought I'd ask! 

(Ah, I just realised after typing the whole thing out that this specific request requires re-generating the whole question on every page load which probably makes this infeasible. I'm still posting it in case anyone has clever ideas to sidestep the question caching process for _just_ this block, or hijack the existing question versioning somehow, etc). 

I frequently want to re-use questions between courses (using a shared question bank), but sometimes need to make minor changes. Any minor change means I need to make a duplicate question, so I end up with a handful of questions like "Implicit partial differentiation (EMTH210)" and "Implicit partial differentiation (MATH201)". 

What I would really love is a block, perhaps something like `[[context]]`, that will check a course ID (or title etc) and only display the relevant information. I'm envisaging something like: 

  `<p>For a similar example see [[context id="EMTH210"]]Section 3.1.2[[id="MATH201"]]Section 2.1.1[[/context]] of the notes.</p>`

I anticipate difficulties with repeating courses - how would I ensure that all iterations of EMTH210 will trigger the relevant id tag? Could there be a central "hub" that works as a "if id='1234' then course='EMTH210'" in the Admin UI? Then you could organise this once per semester. 

If none of the IDs match I would suggest that either nothing displays, or a warning message to contact your teacher displays.

Other places where this would be useful would be default marks or penalties. I suspect this would be more difficult - do these fields ever touch the block system at all? It would be useful to me because some of our courses prefer a short-quiz high-penalty model, with others preferring a long-quiz low-to-none-penalty model. Then I could use the same question in MATH101 with a 10% penalty and in MATH102 with a 50% penalty. If nothing matches here then I think an error needs to be thrown that prevents saving the question. 

Another use case could be defining notation or display standards. For example, a math course might prefer Liebniz notation for a differentiation question whilst a physics course might prefer function notation or dot notation, but otherwise be happy with the exact same question. One course might use a certain colour scheme in JSXGraph to denote some meaning, whereas another course might prefer a different one. 

In the absence of this being a possibility, I will happily accept any suggestions for how best to notate/identify questions like this in a question bank/quiz. At present I'm considering something a little unwieldy like titling my questions ` E210/3. Implicit partial differentiation` and `M201/2. Implicit partial differentiation` to denote course and week number, and ensuring they're in the same/nearby folders. 

Thanks!

sangwinc:
[...]

PR #1852, opened 2026-08-25
Allow semantic wrappers around matrix TeX

adamant-pwn:
### Summary

- Extend `stack_matrix_disp(m)` with an optional TeX wrapper argument.
- With `stack_matrix_disp(m, "\\stackmatrix")`, emit `\stackmatrix{left}{right}{array}`.
- Carry the question's configured delimiters as the first two arguments and use dots for invisible delimiters.
- Preserve the existing one-argument output exactly.

### Motivation

This provides a small platform-neutral hook for late-bound notation. A wrapper can use the first two arguments as the question's configured fallback or ignore them in favour of another presentation. Different wrappers can distinguish semantic matrices, column vectors and row vectors without making STACK inspect Moodle course context or changing compilation, grading or input syntax.

This PR only changes the TeX emitted when an author opts in; it does not install or order wrapper definitions. The ordinary one-argument call remains unchanged.

The corresponding HTML-input distinction remains in #1848. A TeX wrapper cannot restyle STACK's ordinary HTML input grid.

The shared-question use case has also been discussed in [#1327](https://github.com/maths/moodle-qtype_stack/issues/1327) and [MathJax #3603](https://github.com/mathjax/MathJax/issues/3603).

### Verification

- Focused Maxima execution covers square, round, curly, vertical-bar and invisible delimiters and confirms unchanged one-argument output.
- Added data-driven CASText tests for all five delimiter values.
- PHP syntax and `git diff --check` pass.

### DIFF
--- a/doc/en/CAS/Matrix.md
+++ b/doc/en/CAS/Matrix.md
@@ -82,6 +82,27 @@ There must be a more elegant way to do this!
 
 ## Display of matrices ## {#matrixparens}
 
+### Semantic TeX wrappers
+
+The optional second argument of `stack_matrix_disp` is a TeX command which wraps the matrix body and its
+configured delimiters.  For example,
+
+    texput(matrix, lambda([m], stack_matrix_disp(m, "\\stackmatrix")));
+
+with square `matrixparens` emits TeX of the form
+
+    \stackmatrix{[}{]}{\begin{array}{cc} ... \end{array}}
+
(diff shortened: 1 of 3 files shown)


adamant-pwn:
Closing this for now. The optional wrapper is only useful for our intended case once STACK can provide a stable default command that may be overridden by the surrounding context. Without that part, this adds author-facing API without delivering the complete mechanism. I’ll retain the prototype and revisit it with a complete proposal.

PR #1853, opened 2026-08-25
Support providecommand in STACK MathJax output

adamant-pwn:
### Summary

- Add LaTeX-compatible `\providecommand` support to STACK's Moodle MathJax output.
- Load the compatibility AMD only when rendered CASText contains `\providecommand`.
- Support MathJax 2, 3 and 4, including MathJax that is already running and content inserted through Moodle fragments.
- Add equivalent support to the minimal API client.

### Motivation

This lets a portable question carry its own fallback notation while allowing an earlier surrounding definition to win. For example, a question can contain

```tex
\providecommand{\vect}[1]{\mathbf{#1}}
```

while a course defines

```tex
\def\vect#1{\vec{#1}}
```

The same question then works with its own default in native LaTeX or an uncustomised STACK page, and adopts the course notation when the earlier definition is present. It does not need an attached script or site-wide MathJax configuration.

This provides the TeX-side foundation for the course-overridable semantic matrix and vector hooks discussed in [#1848](https://github.com/maths/moodle-qtype_stack/issues/1848#issuecomment-5412139975). It does not itself add those wrappers or change the HTML matrix-input widget; those remain separate follow-up work.

MathJax does not currently implement `\providecommand`; the MathJax maintainer supplied the parser-level workaround used here in [MathJax #3603](https://github.com/mathjax/MathJax/issues/3603#issuecomment-5372267553).

### Implementation notes

The command reuses MathJax's existing `newcommand` parser utilities and preserves an existing macro or delimiter definition. The Moodle integration handles both startup-time configuration and a running MathJax instance. If fragment content was typeset first, only STACK equations containing a `providecommand` error are reprocessed. Ordinary STACK mathematics does not load the module.

The minimal API client has no Moodle AMD loader, so its fixed MathJax 3 setup registers the same small command map directly. This is a separate commit so it can be reviewed independently.

The implementation uses MathJax internals, following the maintainer's proposed workaround, and therefore needs retesting when STACK changes supported MathJax versions. It can be removed when MathJax provides the command natively.

This PR only supplies the generic TeX primitive. It does not add course-aware logic to STACK and does not change STACK's ordinary HTML input widgets.

### Verification

- STACK Jest suite: 33 suites and 445 tests passed.
- Official MathJax 2.7.9, 3.2.2 and 4.0.0 packages: fallback definitions, earlier overrides, cross-expression persistence and optional default arguments passed.
- Moodle 5.1.6 with STACK, PostgreSQL and goemaxima: ordinary page, authenticated AJAX fragment, no-override fallback, mixed square/round fallbacks and ordinary-math control passed.
- MathJax 2.7.9: Moodle 4.2-style delayed startup and late `noErrors` fragment repair passed.
- Minimal API header: actual-browser fallback and earlier-override test passed on MathJax 3.2.2.
- The same fallback/override fixture compiled with pdfLaTeX.


adamant-pwn:
One broader motivation worth recording is that this offers a narrower alternative for the notation-only part of #993 and #1327. Those issues considered making STACK read course or category context, which raised valid concerns about caching, regrading and platform independence; #1327 therefore pointed towards gitsync-managed course variants. `\providecommand` keeps the dependency in the other direction: a self-contained question—and potentially STACK's own matrix, row-vector and column-vector TeX output—can expose semantic commands with portable defaults, while the surrounding rendering context may override those commands. STACK still does not need to know which course it is in, and input or grading semantics are unchanged. This does not replace gitsync for real content or behaviour differences, but it may avoid separate course branches when the only difference is notation.

sangwinc:
Thanks @adamant-pwn I'm going to need a little time to digest and consider this.  Any comments @timhunt and @aharjula in the mean time?

adamant-pwn:
Update: MathJax has now implemented native `\providecommand` in [mathjax/MathJax-src#1555](https://github.com/mathjax/MathJax-src/pull/1555), targeting 4.2. It uses the same rule as this PR: parse through `newcommand` and add the definition only when neither a macro nor delimiter already exists, which seems to validate the approach here.

The compatibility condition is feature-based rather than a permanent version split: if the active TeX input already recognises `\providecommand`, STACK should leave it alone; otherwise it installs the fallback. The implementation still needs separate adapters because MathJax 2 extends `Hub`/`InputJax`/`Parse`, while MathJax 3 and 4 use `CommandMap`/`Configuration`; the new `NewCommand(..., provide)` hook itself only exists in 4.2.

[Moodle main currently defaults to MathJax 4.0.0](https://github.com/moodle/moodle/blob/main/public/filter/mathjaxloader/settings.php), so the fallback remains useful for current installations. Once native support is available throughout STACK's supported Moodle environments, the compatibility AMD can be skipped and eventually removed.

PR #1858, opened 2026-09-07
asciimath-update - Update to version 2.5.1 of ASCIIMath (May 2026)

EJMFarrow:


sangwinc:
Thanks @EJMFarrow, I've tested this locally, and it adds some welcome operators to asciimath such as `mgt` which was missing before.

PR #1860, opened 2026-09-08
Tests fix

EJMFarrow:
- Set a default the absence of which was throwing a PHPUnit warning in Moodle 5.3+.
- Deal in Behat tests with an added PRT now being open by default in some (but not all!) versions of Moodle.


sangwinc:
Thanks @EJMFarrow!

Issue #1850, opened 2026-08-25
Supporting augmented matrices directly in the fixed matrix input


sangwinc:
This is to separate out this suggestion from issue #1848 , as suggested by @adamant-pwn

> One adjacent question prompted by `aug`: would there be any interest in supporting augmented matrices directly in the fixed matrix input? For example, an instantiated `aug(A,b)` model answer could render as one grid with an internal separator, while `aug(A,B,C)` could produce several blocks. Since `aug_matrix` currently lives in `contrib/matrix.mac`, I have not included this in [#1847](https://github.com/maths/moodle-qtype_stack/pull/1847). I am not sure how much demand there would be for this kind of input, but I would be happy to look into it if others would find it useful.

adamant-pwn:
Just today, I was editing a question and thought "Hm, it would be nice to show this system in row-echelon form", so I think in principle it is quite meaningful to allow such inputs to e.g. check results of forward propagation in Gauss algorithm. Happy to suggest some implementation if there is interest in having this option.

rmueller-upb:
Thanks for your initiative, I love this idea. When working with Jordan Normal Forms or Frobenius Normal Forms, this also came to my mind.

sangwinc:
Thanks for pushing these changes to the issue branch.

Could I please ask, @adamant-pwn, if you're willing to improve the validation feedback for variable matrix size?

Here is an example:

<img width="1097" height="208" alt="Image" src="https://github.com/user-attachments/assets/a0b0f938-31a9-44ed-918d-f31efb071eec" />

The validation error message is helpful, but not at all specific.    Could we take one row of the teacher's answer, and use this to format an additional sentence:  

    "Your matrix needs to have the following structure on each row:  `? ? | ?`."

Otherwise there's no (general) way for students to know what to do!  Yes, this gives away the size needed somewhat (one of the purposes of varmatrix is to let us assess students' knowledge of the matrix size!).  We have a compromise between the following

1. Telling the student what size and shape is needed.
2. Not being able to get a valid expression.  

An alternative is to give away the structure.

    "Your matrix needs to have the following structure on each row:  `M | c`, where "M" is a matrix of any size, and "c" is a column vector."

In this example we have the teachers answer built from `aug(M,c)`, and we get the following 

<img width="753" height="219" alt="Image" src="https://github.com/user-attachments/assets/c0af9c7d-4f4b-40ee-8e53-20d8da490199" />

[...]

PR #1861, opened 2026-09-08
Preserve augmented values in fixed and variable-size matrix inputs

adamant-pwn:
Follow-up to #1859, based on the maintainer integration branch `iss1850`. Replaces the explicit separator option with native augmented model-answer support.

Fixed and variable-size matrix inputs now accept an augmented model answer directly, for example `ta:aug(A,b)`. The instantiated `aug_matrix(...)` value determines the grid size and block boundaries. Students fill one grid; submission, saved responses, syntax hints and AJAX validation preserve the augmented constructor and its matrix/column-vector/row-vector blocks. The validation echo renders the augmented matrix with its separators.

This follows the value-preserving approach used for row and column vectors in #1847/#1848 and addresses #1850. Multiple blocks such as `aug(A,B,C)` are supported. Blocks must be nonempty and have equal row counts. There is no separate boundary option.

The existing `aug`, `aug_matrix` TeX rules and `de_aug` helpers move from the contributed matrix library into the linear-algebra core so ordinary validation sessions can render the same value. Vector blocks are converted to bare matrices only for display and `de_aug`, retaining their constructors in the submitted augmented object. Matrix-specific answer tests can use `de_aug(ans1)` explicitly.

API consumers receive the augmented value type and a `blocks` array with each block's constructor and column count. Ordinary matrix input fields, response serialization and API shape remain unchanged.

Validation: isolated STACK API/Maxima checks cover matrix, column-vector, row-vector and multiple-block roundtrips; automatic boundaries; readonly grids; native and AJAX validation echoes; malformed shapes; partial blanks; forbidden cell functions; and the core helpers. Added Moodle PHPUnit and Maxima test cases. The complete Moodle PHPUnit suite was not available locally.

### Screenshots

Captured from the patched STACK input renderer and its actual CAS validation output in an isolated API harness, with MathJax 3. These are not screenshots of an ETH Moodle deployment.

Column-vector augmented block and validation echo

Three augmented blocks and validation echo

### CI context

The earlier separator-only commit's [run 34235575242](https://github.com/maths/moodle-qtype_stack/actions/runs/34235575242) had two passing jobs and three failures at `create_prt_node.feature:75` (`Node 2` found but not visible). The same failure also occurs in the independent upstream [dev run 34235897023](https://github.com/maths/moodle-qtype_stack/actions/runs/34235897023). This does not establish the current native extension's CI result; its own checks are still running.

### Maintainer sample follow-up

Updated `samplequestions/stacklibrary/Topics/LinearAlgebra/LinearSystems/Linear-system.xml` as requested in review. The input model is `aug(TAM1)` with automatic boundaries; `columnseparators` and the contributed-library include are removed. PRTs explicitly flatten `ans1` with `de_aug` for matrix operations, while feedback displays the preserved augmented object.

The existing correct-answer and consequential-marking tests now submit augmented values. A third case rejects an equivalent but unreduced augmented system. All three cases pass on all six deployed seeds: 18/18 tests using the patched STACK API and Maxima runtime. The maintainer's latest `iss1850` changes, including the `setdiagmx` variable-name fix, are preserved.

### Variable-size augmented input

`varmatrix` now preserves augmented answers as well. Students type spaces between entries, newlines between rows and `|` between blocks on every row. Matrix dimensions remain student-selected; the model fixes block count and constructor types only. Column-vector blocks remain one column and row-vector blocks remain one row. Inconsistent separators, empty blocks and malformed cells produce validation errors.

This reuses the previously accepted varmatrix row/column-vector handling from #1847 (which is on `iss1848`, not yet on this PR's base) and extends it for augmented blocks. It preserves ordinary matrix syntax and exposes augmented input metadata to API/CSS consumers. All four standard outer-bracket options work for the textarea and validation. A separate semantic CSS wrapper for augmented MathJax output is not part of this change.

Validation: 56 isolated runtime assertions pass, including resized dimensions, matrix/vector compatibility, multiple blocks, syntax hints, readonly display, decimal commas, implied multiplication, allowed empty answers, malformed inputs, AJAX and forbidden cell functions. The linear-system sample also passes all 18 tests with its first input changed to varmatrix, preserving the follow-through checks. Moodle PHPUnit cases are added; the complete suite remains an upstream CI check.

Variable-size augmented input and validation echo

### CI refresh

Merged current upstream `dev` (including #1860) into this PR as `5f2d0b8d8`. Upstream [run 34247459194](https://github.com/maths/moodle-qtype_stack/actions/runs/34247459194) passes with the PRT-node visibility fix and missing library-import default. Of the five jobs in our older completed run 34242042739, one passed, three failed at that Node 2 step, and the Moodle 4.2 job hit browser connection/time-out errors. Fresh CI on the merged revision must verify the result; those browser failures are not assumed to be resolved.

### DIFF
--- a/classes/library_import.php
+++ b/classes/library_import.php
@@ -346,7 +346,7 @@ class library_import extends \external_api {
         $moduleinfo->navmethod = $quizdata->quiz->navmethod;
         $moduleinfo->timeopen = 0;
         $moduleinfo->timeclose = 0;
-        $moduleinfo->duedate = 0; // Prevent "Undefined property" warning in Moodle 5.3+.
+        $moduleinfo->duedate = null; // Prevent "Undefined property" warning in Moodle 5.3+.
         $moduleinfo->decimalpoints = 2;
         $moduleinfo->questiondecimalpoints = -1;
         $moduleinfo->grademethod = 1;
--- a/classes/library_import.php
+++ b/classes/library_import.php
@@ -346,6 +346,7 @@ class library_import extends \external_api {
         $moduleinfo->navmethod = $quizdata->quiz->navmethod;
         $moduleinfo->timeopen = 0;
         $moduleinfo->timeclose = 0;
+        $moduleinfo->duedate = 0; // Prevent "Undefined property" warning in Moodle 5.3+.
         $moduleinfo->decimalpoints = 2;
         $moduleinfo->questiondecimalpoints = -1;
         $moduleinfo->grademethod = 1;
--- a/classes/library_import.php
+++ b/classes/library_import.php
@@ -346,7 +346,7 @@ class library_import extends \external_api {
         $moduleinfo->navmethod = $quizdata->quiz->navmethod;
         $moduleinfo->timeopen = 0;
         $moduleinfo->timeclose = 0;
-        $moduleinfo->duedate = null; // Prevent "Undefined property" warning in Moodle 5.3+.
+        $moduleinfo->duedate = 0; // Prevent "Undefined property" warning in Moodle 5.3+.
         $moduleinfo->decimalpoints = 2;
         $moduleinfo->questiondecimalpoints = -1;
         $moduleinfo->grademethod = 1;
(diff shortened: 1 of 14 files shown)


sangwinc:
Thanks @adamant-pwn that's very sensible. 

As you can see from my previous commit d58cb0bace723ab54df1320a090fc8409f13005f I adapted one of the sample questions.  I also noticed that you need `aug` from the matrix library (!) and moving this to linear algebra core is very sensible, thanks.  Could I ask you to review and update that sample question in this pull request?

adamant-pwn:
Hi @sangwinc, thanks a lot! I now updated the sample question to follow this PR and use native augmented input.

sangwinc:
Thanks @adamant-pwn that's a substantial improvement and much appreciated.

Issue #1781, opened 2026-06-25
Dropdown input-type requires validation blocks to exists.


eerohirvijoki:
Leaving [[validation:ans_1]] block out, and using the dropdown input type together with the options mustverify and showvalidation set to 0, causes an error during importing the question to moodle. This should probably be fixed?

    <input>                                                                                                                              
      <name>ans_1</name>                                                                                                      
      <type>dropdown</type>                                                                                                              
      <tans>ta_1</tans>                                                                                                       
      <boxsize>15</boxsize>                                                                                                              
      <strictsyntax>1</strictsyntax>                                                                                                     
      <insertstars>0</insertstars>                                                                                                       
      <syntaxhint></syntaxhint>                                                                                                          
      <syntaxattribute>0</syntaxattribute>                                                                                               
      <forbidwords></forbidwords>                                                                                                        
      <allowwords></allowwords>                                                                                                          
      <forbidfloat>1</forbidfloat>                                                                                                       
      <requirelowestterms>0</requirelowestterms>                                                                                         
      <checkanswertype>0</checkanswertype>                                                                                               
      <mustverify>0</mustverify>                                                                                                         
      <showvalidation>0</showvalidation>                                                                                                 
      <options></options>                                                                                                                
    </input>                          

The attached file should function as an example. Trying to import it should produce the message:

questiontext: The question text must contain the token '[[validation:ans_suutin_eta_N]]'. The question text must contain the token '[[validation:ans_suutin_eta_pN]]'. The question text must contain the token '[[validation:ans_diff_eta_D]]'. The question text must contain the token '[[validation:ans_diff_eta_pD]]'. The question text must contain the token '[[validation:ans_suutin]]'. The question text must contain the token '[[validation:ans_diff]]'. The question text must contain the token '[[validation:ans_patotila]]'. The question text must contain the token '[[validation:ans_pitot]]'.
The question has been marked as broken and will not be accessible to students.

[suutin-maaritelmat-2p.xml](https://github.com/user-attachments/files/29330393/suutin-maaritelmat-2p.xml)

sangwinc:
Thanks for contacting us @eerohirvijoki 

By design, every input _must_ have a corresponding validation block.

The situation where you have a (1) a dropdown input, and (2) don't show the validation does not make use of this block.  Other inputs still use the block for error messages, but (hopefully!) the dropdown can't generate errors.

[...]

PR #1862, opened 2026-09-09
Test XML import and editor validation consistency

adamant-pwn:
XML imports should retain the same authoring diagnostics as the normal editor's shared `validate_fromform` path. This adds regression coverage for missing validation placeholders across all 18 installed input types, seven other malformed-definition cases, and valid algebraic/dropdown/matrix controls. Invalid imports must contain the corresponding diagnostic and be marked broken; valid controls must remain unbroken.

This is test-only: the current implementation already passes these cases. In particular, it does not change the deliberate dropdown-placeholder policy discussed in #1781, and it does not treat retaining an invalid question marked broken as silently accepting it.

Validation: Moodle 5.0.9+, PHP 8.3.33, PostgreSQL 17.11, matching `2026080600` STACK CAS libraries. The 28 new cases plus six existing XML-import tests pass: **34 tests, 225 assertions**. PHPUnit reports two metadata deprecations. The full Moodle matrix and UI save workflow were not run.

### DIFF
--- a/tests/questiontype_test.php
+++ b/tests/questiontype_test.php
@@ -862,6 +862,132 @@ final class questiontype_test extends qtype_stack_walkthrough_test_base {
         ));
     }
 
+    /**
+     * XML import must flag the same invalid definitions as the editor's shared validator.
+     *
+     * @dataProvider import_validation_cases
+     * @param string $type Input type.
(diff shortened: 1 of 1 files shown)


adamant-pwn:
Moved to https://github.com/maths/moodle-qbank_importasversion/pull/26.

PR #1864, opened 2026-09-09
Clarify required block structure in variable-matrix validation

adamant-pwn:
When a variable-size augmented matrix has missing separators or incompatible block shapes, validation now shows a short multi-line example, such as:

```text
1 ... | 2
3 ... | 4
```

Row-vector models still show only one row. Dots indicate optional extra entries in matrix and row-vector blocks; column-vector blocks show one entry. A short reminder explicitly leaves the row count to the student while requiring consistent block widths, or states the single-row requirement. The example uses no teacher entries or dimensions, and the existing structural restrictions are unchanged.

Addresses the [maintainer request in #1850](https://github.com/maths/moodle-qtype_stack/issues/1850#issuecomment-5600491744). Based on the updated `iss1850` branch, including the code tidy at `b1d92f882`.

### Screenshots

Captured from STACK's actual `render` and `render_validation` output, using identical responses before and after, in an isolated Moodle test harness with MathJax 3. The surrounding comparison layout is illustrative; these are not screenshots of an ETH Moodle deployment.

**Missing separators for a matrix and column vector:**

Before and after: missing separators, with example 1 ... | 2

**Three blocks, with too many entries in the final column-vector block:**

Before and after: three blocks, with example 1 ... | 2 ... | 3

### Validation

- Variable-matrix PHPUnit suite: 41 tests, 280 assertions pass.
- Fixed-matrix PHPUnit suite: 31 tests, 118 assertions pass.
- All ten new regression cases fail without the PHP fix. Patched cases also check AJAX validation, column-only blocks, different teacher dimensions, and absence of teacher entries in the hint.
- Screenshot capture runs pass against both original and patched renderers: one test / four assertions each.
- PHP syntax checks and `git diff --check` pass.

Tested with Moodle 5.0.9+, PHP 8.3.33, PHPUnit 11.5.55, and Maxima 5.44.0 using the issue branch's STACK-Maxima libraries (`2026080700`). The isolated CAS adapter invokes a fresh Maxima process per request. The final clean-checkout runs report only PHPUnit metadata deprecation notices. Full upstream CI remains pending.


sangwinc:
Thanks @adamant-pwn I'll review that tomorrow.

PR #1866, opened 2026-09-10
Add metadata

EJMFarrow:
- Add metadata to STACK library questions.
- Update metadata docs with more guidance.
- Add label to language input instructing to use ISO codes.
- Looks like I've updated some line endings on a few files (recently updated Qs?) as well in the process - e.g. samplequestions/stacklibrary/Algebra-Refresher/03-Addition-and-subtraction-of-fractions/AlgMap-3-1.xml Seems fine, though.

This adds CC-SA-4.0 as the licence (Moodle code for CC BY-NC as all CC licences seem to be BY https://creativecommons.org/cc-licenses/) for all STACK library questions apart from HELM where we use CC-NC-4.0.


sangwinc:
Thanks @EJMFarrow

